PR #4, opened 2024-04-29
feature: add ZREMRANGEBYSCORE with additional limit

szuecs:
feature: add ZREMRANGEBYSCORE with additional limit

context: go-redis https://pkg.go.dev/github.com/go-redis/redis\#Client.ZRangeByScoreWithScores has this, which we use in https://github.com/zalando/skipper/blob/678fc5a604483a527e6b69645f497675e020328c/net/redisclient.go\#L431-L448 to implement one part of a timewindow based rate limit in https://github.com/zalando/skipper/blob/678fc5a604483a527e6b69645f497675e020328c/ratelimit/redis.go\#L235


rueian:
Hi @szuecs,

Thank you for sharing your use case. However, your use case is ZRANGEBYSCORE, not ZREMRANGEBYSCORE.

It looks like [ZREMRANGEBYSCORE](https://redis.io/docs/latest/commands/zremrangebyscore/) doesn't support the WITHSCORES and LIMIT options you are going to add.

szuecs:
@rueian oops, yes true.

PR #21, opened 2024-12-29
feat(url): Add valkey and valkeys schemes

jouir:
Inspired by the valkey-py client.

https://github.com/valkey-io/valkey-py/blob/v6.0.2/valkey/_parsers/url_parser.py#L40

### DIFF
--- a/url_test.go
+++ b/url_test.go
@@ -15,7 +15,13 @@ func TestParseURL(t *testing.T) {
 	if opt, err := ParseURL("rediss://"); err != nil || opt.TLSConfig == nil {
 		t.Fatalf("unexpected %v %v", opt, err)
 	}
+	if opt, err := ParseURL("valkeys://"); err != nil || opt.TLSConfig == nil {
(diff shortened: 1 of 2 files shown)


jouir:
Added tests

url_test.go
```diff
@@ -15,12 +15,18 @@ func TestParseURL(t *testing.T) {
 	if opt, err := ParseURL("rediss://"); err != nil || opt.TLSConfig == nil {
 		t.Fatalf("unexpected %v %v", opt, err)
 	}
+	if opt, err := ParseURL("valkeyss://"); err != nil || opt.TLSConfig == nil {
```
rueian: Should this be valkeys?

jouir: You are right! It's fixed. Thanks.

jouir:
I've also removed the test on the init address for the valkey scheme because the goal on this line is to test the scheme, not the init address that is already covered by other tests.

PR #22, opened 2024-12-29
feat(url): Add skip_verify param

jouir:
Add `skip_verify` or `skip_verify=true` to disable TLS certificate verification. Default is false.

Inspired by various Go drivers:
- ClickHouse: https://github.com/ClickHouse/clickhouse-go/blob/v2.30.0/clickhouse_options.go#L259
- MongoDB: https://github.com/mongodb/mongo-go-driver/blob/v2.0.0/x/mongo/driver/connstring/connstring.go#L609
- MySQL: https://github.com/go-sql-driver/mysql/blob/v1.8.1/dsn.go#L175

### DIFF
--- a/url.go
+++ b/url.go
@@ -89,6 +89,18 @@ func ParseURL(str string) (opt ClientOption, err error) {
 		_, addr = parseAddr(addr)
 		opt.InitAddress = append(opt.InitAddress, addr)
 	}
+	if opt.TLSConfig != nil && q.Has("skip_verify") {
+		skipVerifyParam := q.Get("skip_verify")
+		if skipVerifyParam == "" {
+			opt.TLSConfig.InsecureSkipVerify = true
+		} else {
+			skipVerify, err := strconv.ParseBool(skipVerifyParam)
+			if err != nil {
+				return opt, fmt.Errorf("valkey: invalid skip verify: %q", skipVerifyParam)
+			}
+			opt.TLSConfig.InsecureSkipVerify = skipVerify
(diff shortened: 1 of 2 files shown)


url.go
```diff
@@ -89,6 +89,17 @@ func ParseURL(str string) (opt ClientOption, err error) {
...
+	if opt.TLSConfig != nil && q.Has("skip_verify") {
+		if q.Get("skip_verify") == "" {
+			opt.TLSConfig.InsecureSkipVerify = true
+		} else {
+			skipVerify, err := strconv.ParseBool(q.Get("skip_verify"))
+            if err != nil {
+				return opt, fmt.Errorf("valkey: invalid skip verify: %q", q.Get("skip_verify"))
+			}
+			opt.TLSConfig.InsecureSkipVerify = skipVerify
+		}
```
rueian:
```suggestion
		if v := q.Get("skip_verify"); v == "" {
			opt.TLSConfig.InsecureSkipVerify = true
		} else {
			skipVerify, err := strconv.ParseBool(v)
			if err != nil {
				return opt, fmt.Errorf("valkey: invalid skip verify: %q", v)
			}
			opt.TLSConfig.InsecureSkipVerify = skipVerify
		}
```

rueian: LGTM mostly, but the indention on line 97 is wrong, and one minor improvement is you can store the result of q.Get() into a variable.

jouir: Suggestions applied

PR #27, opened 2025-01-23
fix #25: Add `IFEQ` option to `SET` command

### LINKED ISSUE #25: Add `IFEQ` option to `SET` command
`IFEQ` option is a new feature in Valkey 8.1.

We should support it by updating the corresponding part of `commands.json` with `oneof` block instead of the old `enum`:
https://github.com/valkey-io/valkey-go/blob/92107ce9280700f370c2d9341ac12212692778d5/hack/cmds/commands.json#L5051-L5059

And re-run `go generate` to generate the new command builder for `SET`.

px86:


hack/cmds/commands.json
```diff
@@ -5050,10 +5050,32 @@
...
+                "name": "ifeq",
+                "type": "pure-token",
+                "token": "IFEQ"
+              },
+              {
+                "name": "comparison-value",
+                "type": "string"
+              }
+            ]
+          }
```
rueian: Hi @px86, there is no need to use the `block` type, just merge the current `ifeq` and `comparison-value` into one `string` type argument.

rueian:
Hi @px86, you need to -s (sign off) for every commit. Your first commit is not signed off, or you can merge your two commits into one signed off commit.

rueian:
Thank you so much @px86!

PR #32, opened 2025-02-10
Prevent client infinite retry

acetaxxxx:
The retry function won't stop when it reaches max retires limit.


rueian:
Hi @acetaxxxx, infinite retry is an intentional default behavior. We could rename the `defaultMaxRetries` to something else to avoid further confusion instead of changing the current behavior.

rueian:
Hi @acetaxxxx, would you like to help us come up with a new name for the variable?

PR #37, opened 2025-03-27
fix ctx deadline  while acquiring connection

srikar-jilugu:


pool.go
```diff
@@ -2,47 +2,111 @@ package valkey
...
+	timerOn     bool
+	poolTimeout time.Duration
 }
 
 func (p *pool) Acquire(ctx context.Context) (v wire) {
 	p.cond.L.Lock()
+
+	poolDeadline := time.Time{}
+	if p.poolTimeout > 0 {
+		poolDeadline = time.Now().Add(p.poolTimeout)
```
rueian: I don’t think we need an additional p.poolTimeout. Using the request ctx should be enough.

pool.go
```diff
@@ -2,47 +2,111 @@ package valkey
...
+		cancel  context.CancelFunc
+	)
+	if !poolDeadline.IsZero() {
+		poolCtx, cancel = context.WithDeadline(context.Background(), poolDeadline)
+		defer cancel()
+
+		go func() {
+			<-poolCtx.Done()
+			if poolCtx.Err() == context.DeadlineExceeded { // signal the pool to stop waiting, only if the poolctx is deadline exceeded
+				p.cond.Signal()
```
rueian:
```suggestion
				p.cond.Broadcast()
```
Using Signal seems incorrect. There is no guarantee that Signal will wake the correct goroutine up.

srikar-jilugu: @rueian I agree that it doesn’t guarantee waking up the right goroutine, but I suspect that calling `Broadcast()` every time the ctx deadline exceeds—especially if it happens frequently—could increase mutex contention, as all waiting goroutines wake up unnecessarily and compete for the lock. wdyt?

rueian: Yes, but I am afraid that we have no choice as long as we still use the sync.Cond. If the standard .Signal() wakes a wrong goroutine, which is very likely, then the timeout just doesn’t work.

pool.go
srikar-jilugu: We can remove it if its not required, but we believe having pool timeout would be very helpful for detecting cases where the pool size is insufficient for an application, as frequent scenarios of all conns being busy would indicate this. Otherwise, when a ctx deadline occurs, we can’t reliably determine whether it’s due to the pool being fully utilized or any other reason like cmd latency.

pool.go
srikar-jilugu: In production throughput, Is it reasonable to assume that if a pool context deadline exceeds, it likely affects multiple simultaneous acquire calls, meaning this particular Wait() will be awakened by one signal or another?

pool.go
rueian: I still think it is not necessary. We can implement a more reliable method to determine if the pool is full by adding some metrics counters.

srikar-jilugu: hey @rueian, I've removed the pool timeout from the PR, but I believe the ctx deadline changes are still necessary to prevent indefinite pool blocking. Could you please review it?

rueian: I think we should not modify the ctx in this way, because we should also cover the error of Canceled.

pool.go
```diff
@@ -39,10 +39,49 @@ type pool struct {
...
+		poolDeadline = ctxDeadline
+	}
+
+	var (
+		poolCtx context.Context
+		cancel  context.CancelFunc
+	)
+	if !poolDeadline.IsZero() {
+		poolCtx, cancel = context.WithDeadline(context.Background(), poolDeadline)
+		defer cancel()
```
rueian: Use a separate context for terminating the new goroutine so that we can cover the case of request ctx cancellation too.

pool.go
```diff
@@ -39,10 +39,49 @@ type pool struct {
...
+
+		go func() {
```
rueian, line 52: Can this goroutine be created only when the pool is full?

pool.go
```diff
@@ -39,10 +39,49 @@ type pool struct {
...
 retry:
-	for len(p.list) == 0 && p.size == p.cap && !p.down {
+	for len(p.list) == 0 && p.size == p.cap && !p.down && ctx.Err() == nil {
 		p.cond.Wait()
 	}
+
+	if ctx.Err() != nil {
+
+		if deadPipe, ok := p.dead.(*pipe); ok {
+			deadPipe.error.Store(&errs{error: ctx.Err()})
```
rueian: The dead pipe is shared. We need to make a new one here instead of mutating the shared dead pipe.

pool.go
srikar-jilugu: would listening to the passed `ctx.Done` as well solve this issue? Then `poolCtx.Done` helps in terminating the go routine when deadline exceeds or when the connection is successfully acquired whereas `ctx.Done` helps in terminating when ctx is cancelled

rueian: We should listen to the passed ctx, but I think the poolCtx can be simpler if it is not a child of the passed ctx. Actually it can just be a `Chan strict{}` and we close it at the end of the Acquire.

Or we still let the poolCtx inherit the passed ctx, but inherit with `WithCancelCause`. Then, we can cancel it with a special error at the end of Acquire to determine if the goroutine has no need to invoke Broadcast.

srikar-jilugu: > we still let the poolCtx inherit the passed ctx, but inherit with WithCancelCause. Then, we can cancel it with a special error at the end of Acquire to determine if the goroutine has no need to invoke Broadcast.

This seems to be a better and simpler approach here as we don't have to handle deadline explicitly

srikar-jilugu: made some changes accordingly, please review

pool.go
```diff
@@ -2,10 +2,14 @@ package valkey
 
 import (
 	"context"
+	"errors"
 	"sync"
 	"time"
 )
 
+// ErrAcquireComplete is a special error used to indicate that the Acquire operation has completed successfully
+var ErrAcquireComplete = errors.New("acquire complete")
```
rueian:
```suggestion
var errAcquireComplete = errors.New("acquire complete")
```

pool.go
```diff
@@ -39,10 +43,33 @@ type pool struct {
 
 func (p *pool) Acquire(ctx context.Context) (v wire) {
 	p.cond.L.Lock()
+
+	// Set up ctx handling when waiting for an available connection
+	if len(p.list) == 0 && p.size == p.cap && !p.down && ctx.Err() == nil {
```
rueian:
```suggestion
	if len(p.list) == 0 && p.size == p.cap && !p.down && ctx.Err() == nil && ctx.Done() != nil {
```

rueian:
Hi @srikar-jilugu, you will also need to pass the DCO check by using `git commit -s`.

srikar-jilugu:
> Hi @srikar-jilugu, you will also need to pass the DCO check by using `git commit -s` to rewrite all your commits.
I have signed-off all the commits, not sure why the last commit is failing the DCO check

srikar-jilugu:
Hi @rueian, all the checks have passed, any other changes required?

PR #48, opened 2025-05-22
Fix XRangeEntry.FieldValues loses insertion order and entries with identical keys #47

### LINKED ISSUE #47: XRangeEntry.FieldValues loses insertion order and entries with identical keys
[`XRangeEntry.FieldValues`](https://pkg.go.dev/github.com/valkey-io/valkey-go@v1.0.60#XRangeEntry.FieldValues) is a `map[string]string`. However, stream entries aren't really hashmaps. Converting the entries to a Go map loses information.

The first difference is that the order of the entries is significant [^1]:

> An entry is composed of a list of field-value pairs. The field-value pairs are stored in the same order they are given by the user. Commands that read the stream, such as XRANGE or XREAD, are guaranteed to return the fields and values exactly in the same order they were added by XADD.

The second difference is that streams preserves entries with the same key [^2]:

```
127.0.0.1:6379> XADD example * foo 1 foo 2 bar 3 bar 4
"1747784186966-0"
127.0.0.1:6379> XRANGE example - +
1) 1) "1747784186966-0"
   2) 1) "foo"
      2) "1"
      3) "foo"
      4) "2"
      5) "bar"
      6) "3"
      7) "bar"
      8) "4"
```

Converting to a Go map loses the insertion order and the duplicate entries.

Ideally it would return something like this instead:

```go
type XRangeEntry struct {
	FieldValues []FieldValue
	ID          string
}

type FieldValue struct {
	Field string
	Value string
}
```

[^1]: https://valkey.io/commands/xadd/
[^2]: https://redis.io/blog/youre-probably-thinking-about-redis-streams-wrong/

atharva-navani16:
Pull Request Description
========================

🐛 Problem
----------

The current XRangeEntry.FieldValues implementation uses map\[string\]string, which causes two significant issues:

1.  **Lost insertion order** - Go maps don't preserve the order that fields were added to Redis stream entries
    
2.  **Duplicate keys overwritten** - If a Redis stream entry contains multiple fields with the same name, only the last value is preserved
    

This is problematic for applications that rely on field ordering or need to handle duplicate field names in Redis streams.

✅ Solution
----------

This PR introduces new slice-based structures that preserve both insertion order and duplicate field handling:

### New Types
```go
type XRangeSlice struct {
    FieldValues []XRangeFieldValue
    ID          string
}
```
```go
type XRangeFieldValue struct {
    Field string
    Value string
}
```

### New Methods

*   AsXRangeSlice() - Convert single entry to ordered format
    
*   AsXRangeSlices() - Convert multiple entries to ordered format
    
*   AsXReadSlices() - Convert XREAD/XREADGROUP responses to ordered format
    

### Helper Methods

*   ToMap() - Convert back to map for convenience (loses duplicates)
    
*   GetValue(field) - Get first value for a field
    
*   GetValues(field) - Get all values for a field (useful for duplicates)
    

🔄 Backward Compatibility
-------------------------

✅ **Fully backward compatible** - All existing methods remain unchanged:

*   XRangeEntry struct unchanged
    
*   AsXRangeEntry() method unchanged
    
*   AsXRange() method unchanged
    
*   AsXRead() method unchanged
    

Users can migrate incrementally or use both APIs side by side.

📋 Changes Made
---------------

*   Added XRangeSlice and XRangeFieldValue struct definitions
    
*   Implemented AsXRangeSlice(), AsXRangeSlices(), AsXReadSlices() methods on ValkeyMessage
    
*   Added corresponding delegation methods on ValkeyResult
    
*   Included helper methods for convenient field access
    
*   Maintained full backward compatibility with existing API
    

🎯 Benefits
-----------

✅ **Preserves insertion order** - Fields maintain their original sequence✅ **Handles duplicate keys** - Multiple fields with same name are preserved✅ **Backward compatible** - Existing code continues to work unchanged✅ **Convenient API** - Helper methods for common operations✅ **Type safe** - Clear separation between field names and values

Closes #47

@rueian

### DIFF
--- a/message.go
+++ b/message.go
@@ -372,6 +372,36 @@ func (r ValkeyResult) AsXRead() (v map[string][]XRangeEntry, err error) {
 	return
 }
 
+// AsXRangeSlice delegates to ValkeyMessage.AsXRangeSlice
+func (r ValkeyResult) AsXRangeSlice() (v XRangeSlice, err error) {
+	if r.err != nil {
+		err = r.err
+	} else {
+		v, err = r.val.AsXRangeSlice()
+	}
+	return
+}
+
+// AsXRangeSlices delegates to ValkeyMessage.AsXRangeSlices
+func (r ValkeyResult) AsXRangeSlices() (v []XRangeSlice, err error) {
+	if r.err != nil {
+		err = r.err
+	} else {
+		v, err = r.val.AsXRangeSlices()
+	}
+	return
+}
+
+// AsXReadSlices delegates to ValkeyMessage.AsXReadSlices
+func (r ValkeyResult) AsXReadSlices() (v map[string][]XRangeSlice, err error) {
+	if r.err != nil {
+		err = r.err
+	} else {
+		v, err = r.val.AsXReadSlices()
+	}
+	return
+}
+
 func (r ValkeyResult) AsLMPop() (v KeyValues, err error) {
 	if r.err != nil {
 		err = r.err
@@ -380,6 +410,7 @@ func (r ValkeyResult) AsLMPop() (v KeyValues, err error) {
 	}
 	return
 }
+
 
 func (r ValkeyResult) AsZMPop() (v KeyZScores, err error) {
 	if r.err != nil {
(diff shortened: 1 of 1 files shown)


message.go
```diff
@@ -1003,6 +1034,145 @@ func (m *ValkeyMessage) AsXRead() (ret map[string][]XRangeEntry, err error) {
 	typ := m.typ
 	return nil, fmt.Errorf("%w: valkey message type %s is not a map/array/set", errParse, typeNames[typ])
 }
+// New slice-based structures that preserve order and handle duplicates
```
rueian:
```suggestion
// New slice-based structures that preserve order and duplicates
```

message.go
```diff
@@ -1003,6 +1034,145 @@ func (m *ValkeyMessage) AsXRead() (ret map[string][]XRangeEntry, err error) {
...
+
+// Helper method to get the first value for a field (most common use case)
+func (x XRangeSlice) GetValue(field string) (string, bool) {
+    for _, fv := range x.FieldValues {
+        if fv.Field == field {
+            return fv.Value, true
+        }
+    }
+    return "", false
+}
```
rueian:
```suggestion
```
We don't need these for now.

rueian:
Hi @atharva-navani16, could you also reference the existing AsXRange AsXRead tests to make this PR 100% test covered?

message.go
```diff
@@ -1003,6 +1034,145 @@ func (m *ValkeyMessage) AsXRead() (ret map[string][]XRangeEntry, err error) {
...
+        if IsValkeyNil(err) {
+            return XRangeSlice{ID: id, FieldValues: nil}, nil
+        }
+        return XRangeSlice{}, err
+    }
+    
+    // Convert pairs to slice (preserving order)
+    if len(fieldArray)%2 != 0 {
+        return XRangeSlice{}, fmt.Errorf("field-values array length must be even, got %d", len(fieldArray))
+    }
```
rueian:
```suggestion
```
This check may not be necessary.

rueian:
Hi @atharva-navani16, any updates for adding tests?

coverage.out
```diff
@@ -0,0 +1,3045 @@
+mode: set
+github.com/valkey-io/valkey-go/binary.go:17.37,19.2 1 1
```
rueian: Hi @atharva-navani16, please do not include coverage.out in the commits.

message.go
```diff
@@ -1003,6 +1034,114 @@ func (m *ValkeyMessage) AsXRead() (ret map[string][]XRangeEntry, err error) {
...
+    
+    // Convert pairs to slice (preserving order)
+    fieldValues := make([]XRangeFieldValue, 0, len(fieldArray)/2)
+    for i := 0; i < len(fieldArray); i += 2 {
+        field := fieldArray[i].string()
+        value := ""
+        if i+1 < len(fieldArray) {
+            value = fieldArray[i+1].string()
+        }
```
rueian:
```suggestion
    for i := 0; i < cap(fieldValues); i++ {
        field := fieldArray[i*2].string()
        value := fieldArray[i*2+1].string()
```

message_test.go
```diff
@@ -2070,3 +2070,633 @@ func TestValkeyMessage(t *testing.T) {
...
+				strmsg('+', "value1"),
+				strmsg('+', "field2"),
+			}),
+		})
+		
+		want := XRangeSlice{
+			ID: "1234567890-4",
+			FieldValues: []XRangeFieldValue{
+				{Field: "field1", Value: "value1"},
+				{Field: "field2", Value: ""},
```
rueian:
```suggestion
```

PR #52, opened 2025-06-10
Ensure deterministic routing in _pickMulti and _pickMultiCache

PingXie:
The `_pickMulti `and `_pickMultiCache` functions previously invoked the `SendToReplicas` callback twice for the same batch of commands: once to count commands per connection and a second time to assign them.

If the `SendToReplicas` implementation was non-deterministic (e.g., stateful, like a round-robin counter), this could lead to inconsistent routing decisions between the two loops. This could potentially cause a panic if a nil connection was chosen on the second pass.

This change refactors both functions to call `SendToReplicas` only once per command. The chosen destination connection for each command is stored in a temporary slice, which is then used for dispatching. This guarantees that the routing decision is consistent for the entire `DoMulti` or `DoMultiCache` operation.

A test case with a stateful `SendToReplicas` function has been added to verify the fix and prevent regressions in both code paths.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .idea/
 dist/
 vendor/
+*.swp
--- a/cluster.go
+++ b/cluster.go
@@ -574,7 +574,8 @@ func (c *clusterClient) _pickMulti(multi []Completed) (retries *connretry, init 
 	count := conncountp.Get(len(c.conns), len(c.conns))
 
 	if !init && c.rslots != nil && c.opt.SendToReplicas != nil {
-		for _, cmd := range multi {
+		destination := make([]conn, len(multi))
+		for i, cmd := range multi {
 			var cc conn
 			if c.opt.SendToReplicas(cmd) {
 				cc = c.rslots[cmd.Slot()]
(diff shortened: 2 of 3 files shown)


PingXie:
> Have you seen panics caused by this? I do see the bug here that can cause panic now:

yeah I did run into this issue. to be exact, it is not `re` but `re.commands` that is nil since map returns an empty object if the key is not found.

> One way to avoid allocations as possible but still only invoking SendToReplicas once per command could be using a bitmap:

good point. let me try it out in my pr next. thanks for the suggestion, @rueian.

PingXie:
> only invoking SendToReplicas once per command could be using a bitmap:

I tried a different approach. I assume that it is a less common case for someone to batch more than 32 commands at once and if that happens the heap allocation is probably not a concern anyways. this option uses some more memory (256 bytes) on the stack but reduces one memory access indirection, comparing to the bitmap solution. wdyt?

rueian (approved):
LGTM. `[32]conn` actually takes 512 bytes. https://go.dev/play/p/uQljjZoV4pc

PingXie:
> LGTM. `[32]conn` actually takes 512 bytes. https://go.dev/play/p/uQljjZoV4pc

Ah you are right. `conn` is actually a value type. thanks @rueian!

Issue #46, opened 2025-05-05
High loading errors during failover of a valkey cluster


srikar-jilugu:
We’re using valkey-go to connect to one of our production Valkey clusters. During a failover triggered as part of an engine upgrade (from valkey 7.x to 8.x in our case), we observed a significant spike in loading errors—something we typically don’t see when using go-redis. Upon investigating further, it appears that valkey-go includes nodes in the cluster topology even when they’re in a `LOADING` state, causing all requests routed to those nodes to fail.

To mitigate this, could we consider temporarily excluding nodes in the `LOADING` state from the topology? go-redis handles this case in a similar way, largely because it determines the routing for each command on the fly. In contrast, since valkey-go refreshes the cluster state asynchronously, implementing such filtering might be a bit more tricky.

rueian:
Hi @srikar-jilugu, if we exclude them, where should we send the command to?

srikar-jilugu:
This is great!, will test this out on our systems, also do we configure somewhere in the client on which command to use for retreving the topology? (cluster slots or shards)

srikar-jilugu:
@rueian this blog says that the `health` field is present in cluster shards even in valkey 7, Can we configure to use cluster shards instead of slots?

rueian:
Unfortunately, CLUSTER SHARDS had a server side bug which made it unusable until Redis 8 and Valkey 7.2.6

See related issues in this PR https://github.com/redis/rueidis/pull/664

PR #56, opened 2025-06-20
[draft] unhealthy node handling

srikar-jilugu:
This is to address: https://github.com/valkey-io/valkey-go/issues/46

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -206,15 +206,28 @@ func (c *clusterClient) _refresh() (err error) {
 	pending = nil
 
 	groups := result.parse(c.opt.TLSConfig != nil)
+
 	conns := make(map[string]connrole, len(groups))
 	for master, g := range groups {
 		conns[master] = connrole{conn: c.connFn(master, c.opt)}
 		if c.rOpt != nil {
 			for _, nodeInfo := range g.nodes[1:] {
+				// do not include unhealhty connections in this refresh cycle
+				if cc, ok := c.conns[nodeInfo.Addr]; ok {
+					if cc.conn.IsServerUnHealthy() {
+						continue
+					}
+				}
 				conns[nodeInfo.Addr] = connrole{conn: c.connFn(nodeInfo.Addr, c.rOpt)}
 			}
 		} else {
 			for _, nodeInfo := range g.nodes[1:] {
+				// do not include unhealhty connections in this refresh cycle
+				if cc, ok := c.conns[nodeInfo.Addr]; ok {
+					if cc.conn.IsServerUnHealthy() {
+						continue
+					}
+				}
 				conns[nodeInfo.Addr] = connrole{conn: c.connFn(nodeInfo.Addr, c.opt)}
 			}
 		}
@@ -524,6 +537,11 @@ process:
 		resp = results.s[1]
 		resultsp.Put(results)
 		goto process
+	case RedirectLoadingRetry:
+		// mark the associated node temporarily unhealthy
+		cc.SetServerUnHealthy()
+		c.refresh(ctx) // on-demand refresh
+		fallthrough
 	case RedirectRetry:
 		if c.retry && cmd.IsReadOnly() {
 			shouldRetry := c.retryHandler.WaitOrSkipRetry(ctx, attempts, cmd, resp.Error())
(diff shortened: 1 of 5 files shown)


cluster.go
```diff
@@ -206,15 +206,28 @@ func (c *clusterClient) _refresh() (err error) {
...
 				conns[nodeInfo.Addr] = connrole{conn: c.connFn(nodeInfo.Addr, c.rOpt)}
 			}
 		} else {
 			for _, nodeInfo := range g.nodes[1:] {
+				// do not include unhealhty connections in this refresh cycle
+				if cc, ok := c.conns[nodeInfo.Addr]; ok {
+					if cc.conn.IsServerUnHealthy() {
+						continue
+					}
+				}
```
rueian: Hi @srikar-jilugu, thanks for the PR. I think the idea is good. A few thoughts:
1. Skipping the `conns[nodeInfo.Addr]` assignments is not the correct place to exclude unhealthy connections in this refresh cycle. The actual places to skip those connections are the assignments of `pslots[i]` and `rslots[i]`.
2. Furthermore, we can't skip unhealthy connections if they are connected to the masters. What we can only do is avoid picking unhealthy connections for replicas. That is, the only two places we can skip those connections are:
https://github.com/valkey-io/valkey-go/blob/ed9f107ed8d7c0350ca464941851786fce2897d3/cluster.go#L255
https://github.com/valkey-io/valkey-go/blob/ed9f107ed8d7c0350ca464941851786fce2897d3/cluster.go#L284
3. I think there should not be a `SetServerUnHealthy` method. It should be done privately in the `conn`.
4. I think there should not be an `UnHealthyNodeInterval` option, and 15s is too long. I guess we need a periodic active check.
5. `IsLoading` could be better than `IsServerUnHealthy`.

srikar-jilugu: 1,2: I believe, this change would only check the health status for the replica nodes, as we are initiating `connrole` for master seperately here. 
https://github.com/valkey-io/valkey-go/blob/ed9f107ed8d7c0350ca464941851786fce2897d3/cluster.go#L213
Also in the case when there is only one replica and it is unhealthy, we are falling back to master while assigning connections to `pslots` and `rslots`. So this shouldn't cause a situation where we are out of conn assignments or ignore a master node for assignment.

3: Any suggestion on how we can  detect that the node is unhealthy, I think, this method would allow us to extend it to other scenarios (for eg: setting it to unhealthy, when multiple retries are failing) as well in future. 
4. Agreed, this need not be exposed
5. As mentioned in the point 3, I think we can use this status in some other cases in future and not just loading. wdyt?

rueian: > So this shouldn't cause a situation where we are out of conn assignments or ignore a master node for assignment.

Yes, but we still have a final mutation for the `conn` map at L236-L245. Please do the unhealthy check after that: https://github.com/valkey-io/valkey-go/blob/ed9f107ed8d7c0350ca464941851786fce2897d3/cluster.go#L236-L245

> Any suggestion on how we can detect that the node is unhealthy, I think, this method would allow us to extend it to other scenarios.

There are two ways to check if a node is loading: check the `loading_eta_seconds` in an `INFO` response, or check each response if a loading error occurs. I don't have a good idea for detecting other unhealthy scenarios. I guess whether a node is unhealthy or not even varies from user to user. Different users may want different behaviors. Could we just use `IsLoading` for now to avoid overloading the term `Unhealthy` until we come up with an API that allows users to mark a node as unhealthy and specify what to do with it?

srikar-jilugu: @rueian updated the load status handling based on your feedback, roughly
1. triggering the load state setting from an initial loading error response from any command, and setting the timeout for this status to be equal to the eta (afaik the node might not respond to any requests during this period)
2. checking the node load status and considering only the active nodes just before pslots and rslots are assigned

srikar-jilugu: will test this out on our workloads as well, before I make this PR open

rueian:
> I throttled the replica’s I/O bandwidth during startup using `systemd-run`. This kept the replica in a loading state for as long as its needed.

I guess the redis main thread may be busy on writing and reading from its disk during the replication sync. I think you can try to make the client timeout long enough. And then capture a tcpdump from the redis side.

Once you have the tcpdump, you can mock a e2e test.

PR #57, opened 2025-06-30
docs: update valkeylock comment

davidlin-tv2:
Updates comment on default value of ExtendInterval in valkeylock to match the actual default value used.

Might be that the proper solution would be to change the actual default value, but the code there was changed more recently than the comment, so seemed like that was what was intended. Also would cause more breakage for users.

Default value seems to be set only on line 64


rueian:
Hi @davidlin-tv2, thanks! please signed off the commit to make the DCO check pass.

rueian (approved):

Issue #36, opened 2025-03-25
Non-blocking initialization


mickeyreiss:
Is there a signature of `valkey.NewClient` which succeeds even if the valkey server is unavailable?

This would match `database/sql`, which provides a connection pool even if the database server is unavailable. If the server remains unavailable when commands are issued, then I would expect them to fail at call time.

For example:
```golang
client, err := valkey.NewClient(valkey.ClientOption{
	InitAddress: []string{"localhost:4567"},
})
if err != nil {
	panic(err) // does not panic, even though nothing is listening on localhost:4567
}
resp := client.Do(context.TODO(), client.B().Get().Key("example-key").Build())
log.Print(resp.Error()) // error is expected here: "dial tcp 127.0.0.1:4567: connect: connection refused"
```

rueian:
Hi @mickeyreiss, we don’t have that signature, unfortunately.

rueian:
See the previous discussion on a potential implementation https://github.com/redis/rueidis/issues/489#issuecomment-1988702413

Basically, I think this can be implemented with a new wrapper of the Client interface embedded with an atomic pointer plus some constraints on the input ClientOption.

PR #58, opened 2025-07-05
Return a singleClient instance even when Dial fails if ForceSingleClient is true

sugymt:
This partially resolves issue #36.

Regarding the ideal implementation mentioned by @rueian in https://github.com/redis/rueidis/issues/489#issuecomment-1988702413, ForceClusterClient doesn't exist yet. However, ForceSingleClient can be implemented in advance.

This behavior is useful to prevent application pods from entering CrashLoopBackoff when Valkey pods haven't yet started within a Kubernetes cluster.

### DIFF
--- a/client.go
+++ b/client.go
@@ -30,10 +30,14 @@ func newSingleClient(opt *ClientOption, prev conn, connFn connFn, retryer retryH
 
 	conn := connFn(opt.InitAddress[0], opt)
 	conn.Override(prev)
-	if err := conn.Dial(); err != nil {
-		return nil, err
+	err := conn.Dial()
+	if err != nil {
+		if !opt.ForceSingleClient {
+			return nil, err
+		}
 	}
-	return newSingleClientWithConn(conn, cmds.NewBuilder(cmds.NoSlot), !opt.DisableRetry, opt.DisableCache, retryer, opt.ConnLifetime > 0), nil
+	// return a singleClient instance even when Dial fails if ForceSingleClient is true.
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -30,8 +30,10 @@ func newSingleClient(opt *ClientOption, prev conn, connFn connFn, retryer retryH
 
 	conn := connFn(opt.InitAddress[0], opt)
 	conn.Override(prev)
-	if err := conn.Dial(); err != nil {
-		return nil, err
+	if !opt.ForceSingleClient {
+		if err := conn.Dial(); err != nil {
+			return nil, err
+		}
```
rueian, line 37: Hi @sugymt, we can't skip the `conn.Dial` since current users still rely on the error. However, I think we are safe to return the error and the client implementation simultaneously.

rueian: Also, we need new tests to make sure that the client implementation can still work after an initial `conn.Dial()` error.

sugymt:
@rueian I've incorporated the feedback. How does this look?

valkey_test.go
```diff
@@ -242,6 +242,54 @@ func TestFallBackSingleClient(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer ln.Close()
+
+	_, port, _ := net.SplitHostPort(ln.Addr().String())
+	client, err := NewClient(ClientOption{
+		InitAddress:       []string{"127.0.0.1:" + port},
+		ForceSingleClient: true,
+		Dialer:            net.Dialer{Timeout: time.Second},
```
rueian:
```suggestion
		Dialer:            net.Dialer{Timeout: time.Second/10},
```

sugymt:
@rueian Fixed and squashed.

rueian (approved):
Thanks @sugymt!

PR #60, opened 2025-07-17
feat: Implement CLIENT CAPA redirect support for standalone client

SoulPancake:
Solves- #39


rueian:
The idea looks good mostly, but needs some changes.

SoulPancake:
Sure pls suggest @rueian

standalone.go
```diff
@@ -53,29 +61,71 @@ func (s *standalone) pick() int {
...
+		redirectConn := s.connFn(addr, &redirectOpt)
+		if err := redirectConn.Dial(); err != nil {
+			// If redirect fails, return the original result
+			return result
+		}
+		defer redirectConn.Close()
+		
+		// Create a temporary client for the redirect
+		redirectClient := newSingleClientWithConn(redirectConn, cmds.NewBuilder(cmds.NoSlot), !s.opt.DisableRetry, s.opt.DisableCache, s.retryer, false)
+		defer redirectClient.Close()
```
rueian: We need a singleflight here and swap the old `s.primary` with the new client.

standalone.go
```diff
@@ -17,9 +17,13 @@ func newStandaloneClient(opt *ClientOption, connFn connFn, retryer retryHandler)
 		return nil, err
 	}
 	s := &standalone{
-		toReplicas: opt.SendToReplicas,
-		primary:    newSingleClientWithConn(p, cmds.NewBuilder(cmds.NoSlot), !opt.DisableRetry, opt.DisableCache, retryer, false),
-		replicas:   make([]*singleClient, len(opt.Standalone.ReplicaAddress)),
+		toReplicas:     opt.SendToReplicas,
+		primary:        newSingleClientWithConn(p, cmds.NewBuilder(cmds.NoSlot), !opt.DisableRetry, opt.DisableCache, retryer, false),
+		replicas:       make([]*singleClient, len(opt.Standalone.ReplicaAddress)),
+		enableRedirect: opt.Standalone.EnableRedirect,
```
rueian, line 24: We need to make sure that when a user sets `opt.Standalone.EnableRedirect`, it will use the `standalone` implementation.

rueian: We also need to make sure that `EnableRedirect` and `ReplicaAddress` can't be used together.

message.go
```diff
@@ -43,6 +43,14 @@ func IsValkeyBusyGroup(err error) bool {
 	return false
 }
 
+// IsValkeyRedirect checks if it is a valkey REDIRECT message.
+func IsValkeyRedirect(err error) (addr string, ok bool) {
```
rueian: Can we try to remove this function? I think a `if ret, yes := IsValkeyErr(err); yes && ret. IsRedirect {` is enough.

standalone.go
```diff
@@ -12,6 +13,11 @@ func newStandaloneClient(opt *ClientOption, connFn connFn, retryer retryHandler)
 	if len(opt.InitAddress) == 0 {
 		return nil, ErrNoAddr
 	}
+	
+	// Validate that EnableRedirect and ReplicaAddress cannot be used together
+	if opt.Standalone.EnableRedirect && len(opt.Standalone.ReplicaAddress) > 0 {
+		return nil, errors.New("EnableRedirect and ReplicaAddress cannot be used together")
+	}
```
rueian: Move this to the outer level.

standalone.go
```diff
@@ -66,23 +72,25 @@ func (s *standalone) handleRedirect(ctx context.Context, cmd Completed, result V
...
+				return result
+			}
+			defer redirectConn.Close()
+			
+			// Create a temporary client for the redirect
+			redirectClient := newSingleClientWithConn(redirectConn, cmds.NewBuilder(cmds.NoSlot), !s.opt.DisableRetry, s.opt.DisableCache, s.retryer, false)
+			defer redirectClient.Close()
+			
+			// Execute the command on the redirect target
+			return redirectClient.Do(ctx, cmd)
```
rueian: We need a singleflight here and swap the s.primary.

rueian: Hi @SoulPancake, any updates?

SoulPancake: Hi @rueian I was OOO, Will get back to this in a day or two

standalone.go
```diff
@@ -53,29 +63,87 @@ func (s *standalone) pick() int {
...
+func (s *standalone) handleRedirect(ctx context.Context, cmd Completed, result ValkeyResult) ValkeyResult {
+	if !s.enableRedirect {
+		return result
+	}
+
+	if ret, yes := IsValkeyErr(result.Error()); yes {
+		if addr, ok := ret.IsRedirect(); ok {
+			// Use singleflight to ensure only one redirect operation happens at a time
+			if err := s.redirectCall.Do(ctx, func() error {
+				return s.redirectToPrimary(addr)
```
rueian: What if an ongoing singleflight mismatches with the `addr` here?

standalone.go
```diff
@@ -53,29 +63,87 @@ func (s *standalone) pick() int {
...
+			}); err != nil {
+				// If redirect fails, return the original result
+				return result
+			}
+
+			// Execute the command on the updated primary
+			return s.primary.Do(ctx, cmd)
```
rueian: What if this returns a redirect error again?

SoulPancake: Do we retry it? @rueian 
Maybe we can take in a configurable retry limit, What do you suggest?

standalone.go
SoulPancake: I think Singleflight redirect should only proceed if the addr matches the latest seen redirect address

standalone.go
rueian: I think there is no need for a configurable retry limit. Usually, we just retry until the context deadline is reached. You can reuse the `retryHandler`.

standalone.go
rueian: Is there a way to know which address is the last seen one?

SoulPancake:
I was trying to see if Copilot could do it :p , but I guess it is just way faster manually, I'll do it tomorrow if there's no rush @rueian

standalone.go
SoulPancake: We can store it in the standalone struct lastRedirectAddr ? @rueian 



What do you suggest

rueian: I now think retries are all we need.

SoulPancake:
I'm sorry with the delay here, I got busy with work, I will close in this weekend @rueian

standalone.go
```diff
@@ -53,29 +64,126 @@ func (s *standalone) pick() int {
...
+	// Handle redirects with retry until context deadline
+	resp = s.handleRedirect(ctx, cmd, resp)
+	if ret, yes := IsValkeyErr(resp.Error()); yes {
+		if _, ok := ret.IsRedirect(); ok && s.enableRedirect {
+			// Use retryHandler to handle multiple redirects with context deadline
+			if s.retryer.WaitOrSkipRetry(ctx, attempts, cmd, resp.Error()) {
+				attempts++
+				goto retry
+			}
+		}
```
rueian: Since we now have the retry structure, I think we don't need `handleRedirect` anymore, or we can largely simplify  `handleRedirect`. The retry structure should look like this:

```suggestion
    if s.enableRedirect {
      if ret, yes := IsValkeyErr(resp.Error()); yes {
		if addr, ok := ret.IsRedirect(); ok {
		    err := s.redirectCall(ctx, func() {return s.redirectToPrimary(adde)})
			// Use retryHandler to handle multiple redirects with context deadline
			if err == nil || s.retryer.WaitOrSkipRetry(ctx, attempts, cmd, resp.Error()) {
				attempts++
				goto retry
			}
		}
```

standalone.go
```diff
@@ -53,29 +64,126 @@ func (s *standalone) pick() int {
...
+			s.lastRedirectAddr = addr
+			
+			// Use singleflight to ensure only one redirect operation happens at a time
+			// Only proceed if the address matches the latest seen redirect address
+			if err := s.redirectCall.Do(ctx, func() error {
+				// Double-check that this is still the latest redirect address
+				if s.lastRedirectAddr == addr {
+					return s.redirectToPrimary(addr)
+				}
+				// If address doesn't match, another redirect happened, skip this one
```
rueian: > If address doesn't match, another redirect happened, skip this one

Hmm, I think this is not correct. Here, you are currently holding the singleflight, which means that the one who updated the `lastRedirectAddr` is waiting for your result. If you skip the `s.redirectToPrimary(addr)`, then no one will do that.

rueian:
@SoulPancake, one more important thing I forgot to mention. The `standalone` client implementation reuses the `singleClient`, which will recycle the incoming commands automatically. We need to use `command.Pin()` to disable this behavior so that we can retry the commands if redirection is enabled.

SoulPancake:
Hi @rueian I am really sorry, very occupied with work and other things.
I will move this to draft state or close it, maybe someone else can continue on this, I'm unsure when I'll have the time.

SoulPancake:
@rueian I have reopened this

Can you have a look at it?

standalone.go
```diff
@@ -53,29 +63,102 @@ func (s *standalone) pick() int {
...
+	if err := redirectConn.Dial(); err != nil {
+		return err
+	}
+
+	// Create a new primary client with the redirect connection
+	newPrimary := newSingleClientWithConn(redirectConn, cmds.NewBuilder(cmds.NoSlot), !s.opt.DisableRetry, s.opt.DisableCache, s.retryer, false)
+
+	// Close the old primary and swap to the new one
+	oldPrimary := s.primary
+	s.primary = newPrimary
```
rueian: Use an atomic pointer here.

standalone.go
```diff
@@ -53,29 +63,102 @@ func (s *standalone) pick() int {
...
+		resp = s.primary.Do(ctx, cmd)
 	}
-	return s.primary.Do(ctx, cmd)
+
+	// Handle redirects with retry until context deadline  
+	if s.enableRedirect {
+		if ret, yes := IsValkeyErr(resp.Error()); yes {
+			if addr, ok := ret.IsRedirect(); ok {
+				// Pin the command to prevent recycling during retries
+				cmd = cmd.Pin()
```
rueian: It is too late to pin here. We should pin the cmd at the beginning if s.enableRedirect is true.

standalone.go
```diff
@@ -53,29 +65,107 @@ func (s *standalone) pick() int {
...
+	}
+
+	// Create a new primary client with the redirect connection
+	newPrimary := newSingleClientWithConn(redirectConn, cmds.NewBuilder(cmds.NoSlot), !s.opt.DisableRetry, s.opt.DisableCache, s.retryer, false)
+
+	// Atomically swap the primary and close the old one
+	oldPrimary := atomic.SwapPointer(
+		(*unsafe.Pointer)(unsafe.Pointer(&s.primary)),
+		unsafe.Pointer(newPrimary),
+	)
```
rueian: Hi @SoulPancake, please change the type of `s.primary` to `atomic.Pointer[singleClient]`

PR #61, opened 2025-07-25
fix(valkeycompat): clear placeholder pipeline error in SliceCmd and BoolSliceCmd

zeeshan-zomato:


jayKabra-Zoman (approved):

rueian:
??

zeeshan-zomato:
@rueian When executing an HMGET pipeline using slicecmd, the error on the result is not automatically cleared after the pipeline is executed. It retains the placeholder error indicating that the pipeline has not been executed yet, even after Exec is called.

valkeycompat/command.go
```diff
@@ -529,6 +529,7 @@ func (cmd *SliceCmd) from(res valkey.ValkeyResult) {
 		}
 	}
 	cmd.SetVal(vals)
+	cmd.SetErr(err)
```
rueian: Hi @zeeshan-zomato, thanks! But I wonder if we should fix this in the Pipeline.Exec instead.

rueian: And TxPipeline.Exec. Just use `rets[i].SetErr(nil)` to clear all responses before `rets[i].from(...)`

zeeshan-zomato: Makes sense, updated

jayKabra-Zoman (approved):

PR #80, opened 2025-09-23
fix redirections that happen for cmds with InitSlot

skif48:
PR with a fix for the [issue](https://github.com/valkey-io/valkey-go/issues/81)

### DIFF
--- a/cluster_test.go
+++ b/cluster_test.go
@@ -6522,7 +6522,7 @@ func TestClusterClientMovedRetry(t *testing.T) {
 						results[i] = newResult(strmsg('-', "MOVED 1 127.0.0.1"), nil)
 					} else {
 						// return a successful transaction result for the retry
-						results[i] = newResult(strmsg('*', "1\r\n+some_value\r\n"), nil)
+						results[i] = newResult(slicemsg('*', []ValkeyMessage{strmsg('+', "some_value")}), nil)
 					}
 				} else {
 					results[i] = newResult(strmsg('+', "QUEUED"), nil)
@@ -6544,6 +6544,10 @@ func TestClusterClientMovedRetry(t *testing.T) {
 
 		if len(resps) != 3 {
 			t.Fatalf("unexpected response length %v", len(resps))
+		}
+
+		if vs, err := resps[2].AsStrSlice(); err != nil || vs[0] != "some_value" {
+			t.Fatalf("unexpected response %v %v", vs, err)
 		}
 	})
 


cluster.go
```diff
@@ -746,7 +746,14 @@ func (c *clusterClient) doresultfn(
 				}
 				retryDelay = c.retryHandler.RetryDelay(attempts, cm, resp.Error())
 			} else {
-				nc = c.redirectOrNew(addr, cc, cm.Slot(), mode)
+				slotToRedirect := cm.Slot()
+				if slotToRedirect == cmds.InitSlot && ei < i {
+					// This block will be triggered in case we've got a MOVED response for a cmd which we assigned cmds.InitSlot to.
+					// For example, it can happen on EXEC within transaction during cluster reshard.
+					// We need to invoke redirectOrNew() with a slot from a previous command.
+					slotToRedirect = commands[i-1].Slot()
```
rueian: Hi @skif48, thanks for reporting the bug and creating this PR. So, we need to find the correct slot for the multi-exec block here, and `commands[i-1]` might not be the correct one we are looking for. It can also be another keyless command, isn't it?

We'd better scan through each command in the multi-exec block to find its slot. However, that is actually optional and might be hard to code. What we must do is modify the `redirectOrNew` function so that it does not do the ad-hoc assignments ([here](https://github.com/valkey-io/valkey-go/blob/9ac1f17dce145b43ba47f3a9df11fc13fd559472/cluster.go#L505-L507) and [here](https://github.com/valkey-io/valkey-go/blob/9ac1f17dce145b43ba47f3a9df11fc13fd559472/cluster.go#L519-L521)) if the slot is `cmds.InitSlot`. Those ad-hoc assignments are purely optimization and can be safely disabled.

skif48: Hey! @rueian

> It can also be another keyless command, isn't it?

Yes, you're totally right, thanks.

> What we must do is modify the redirectOrNew function so that it does not do the ad-hoc assignments ([here](https://github.com/valkey-io/valkey-go/blob/9ac1f17dce145b43ba47f3a9df11fc13fd559472/cluster.go#L505-L507) and [here](https://github.com/valkey-io/valkey-go/blob/9ac1f17dce145b43ba47f3a9df11fc13fd559472/cluster.go#L519-L521)) if the slot is cmds.InitSlot.

This was the first option that I considered, but I wasn't sure wether those two assignments are optional or required. I can see that no matter if we call `redirectOrNew`, the transaction will be retried later, but can you please help me understand which connection it will use from `wslots`? To me it looks it will use the same connection, unless some other command accessed the same slot, got a legit moved for it and called `redirectOrNew` again.

We can also rely on interval based cluster refreshes, but they are optional and can be disabled.

So if neither the slot was accessed by some other command, nor cluster refresh updated connections to slots assignments, we will rely on the fact that the next time we execute this transaction, we're supposed to receive MOVED sooner than on EXEC on a non-keyless command and that will reassign the connection per slot.

Hope my understanding is correct, please help me to make sure. 🙂

rueian: For the immediate retry on the transaction, it will use the connection that connects to the address in the MOVED response.

For `wslots`, whenever we receive a `MOVED ` response, the `shouldRefreshRetry` function triggers `lazyRefresh` and then updates `wslots`. So `wslots` should eventually have the latest cluster topology after receiving a `MOVED`.

skif48: Thanks for the explanation.

I changed the implementation and pushed a test for it, please take a look.

One note about the test though, is that seems like due to how test fixtures are created, my test always goes to [this](https://github.com/valkey-io/valkey-go/blob/9ac1f17dce145b43ba47f3a9df11fc13fd559472/cluster.go#L501) if statement, because `cc.conn` is always nil. I tried to add a test for the scenario where we receive MOVED redirection to the same host, but honestly speaking it looks like it's just not worth the trouble so please let me know what you think.

Also not sure why this test runs for 6 seconds, looks like there's a time.Sleep somewhere. At first I thought it's [this one](https://github.com/valkey-io/valkey-go/blob/9ac1f17dce145b43ba47f3a9df11fc13fd559472/cluster.go#L513), but like I said we do not fall into the `else if` here.

skif48: Also do you want me to open the same PR to https://github.com/redis/rueidis ?

rueian: No need. I will sync your commit to there.

rueian: > Also not sure why this test runs for 6 seconds

No worry. It is caused by the goroutine leak detection.

rueian:
Thanks so much @skif48

skif48:
@rueian hello. Could you please tell when are you planning to publish a new release tag with this fix?

PR #90, opened 2025-11-22
Adding unix socket connection examples

vipoxin:
This pull request adds 2 examples on how to connect to unix socket wit or without helper functions.


vipoxin:
My bad, pasted the example in the wrong place

PR #92, opened 2025-11-26
feat: Add MaxMovedRedirections option to prevent infinite redirect loops

nirchetrit:
## Description
This PR introduces a new `ClusterOption` called `MaxMovedRedirections`. This option allows users to configure a maximum limit for `MOVED` | `ASK` redirections during cluster operations.

Previously, the client could potentially enter an infinite loop of redirects if the cluster was misconfigured or unstable. This change ensures the client fails fast with a specific error (`ErrMaxMovedRedirectionsExceeded`) once the configured redirection limit is reached.

## Changes
- **New Option**: Added `MaxMovedRedirections` to `ClusterOption`.
- **Logic Update**: Updated the retry loops in `cluster.go` (including `DoMulti` and pipelines) to respect this limit.
- **Error Handling**: Enhanced error inspection to ensure we only count actual `MOVED` | `ASK` responses against this limit, preserving other errors (like network issues) where appropriate.

## Motivation
To allow the client to fail faster with a specific error in case of a moved redirection loop during cluster rebalance/topology changes.


cluster.go
```diff
@@ -1259,7 +1297,24 @@ retry:
...
 		if retries.Redirects > 0 {
+			if retries.MovedRetries > 0 {
+				movedRetries++
+				if c.opt.ClusterOption.MaxMovedRedirections > 0 && movedRetries > c.opt.ClusterOption.MaxMovedRedirections {
+					// Set error for all MOVED responses
+					for i := range results.s {
+						if err := results.s[i].val.Error(); err != nil {
+							if vErr, ok := err.(*ValkeyError); ok {
+								if _, isMoved := vErr.IsMoved(); isMoved {
+									results.s[i] = newErrResult(ErrMaxMovedRedirectionsExceeded)
```
rueian: Overall looks good to me, but why don't we just keep the original errors? I am not sure if hiding the redirect information is a good idea.

And should we also apply this configuration to ASK redirection as well?

nirchetrit: I changed it to return the last received redirect error (we now include both ask and moved in the counter)

rueian (approved):

rueian:
Thank you @nirchetrit!

PR #94, opened 2025-12-12
Fix `valkey.dial` span attributes

austindrenski:
From the [OTel semantic conventions](https://opentelemetry.io/docs/specs/semconv/registry/attributes/server/) docs:

> # Server
> 
> ## Server Attributes
> 
> These attributes may be used to describe the server in a connection-based network interaction where there is one side that initiates the connection (the client is the side that initiates the connection). This covers all TCP network interactions since TCP is connection-based and one side initiates the connection (an exception is made for peer-to-peer communication over TCP where the "user-facing" surface of the protocol / API doesn't expose a clear notion of client and server). This also covers UDP network interactions where one side initiates the interaction, e.g. QUIC (HTTP/3) and DNS.
> 
> **Attributes:**
> 
> | Key | Stability | Value Type | Description | Example Values |
> | --- | --- | --- | --- | --- |
> | <a id="server-address" href="#server-address">`server.address`</a> | Stable | string | Server domain name if available without reverse DNS lookup; otherwise, IP address or Unix domain socket name. [1] | `example.com`; `10.1.2.80`; `/tmp/my.sock` |
> | <a id="server-port" href="#server-port">`server.port`</a> | Stable | int | Server port number. [2] | `80`; `8080`; `443` |
> 
> **[1] `server.address`:** When observed from the client side, and when communicating through an intermediary, `server.address` SHOULD represent the server address behind any intermediaries, for example proxies, if it's available.
> 
> **[2] `server.port`:** When observed from the client side, and when communicating through an intermediary, `server.port` SHOULD represent the server port behind any intermediaries, for example proxies, if it's available.

### DIFF
--- a/valkeyotel/metrics.go
+++ b/valkeyotel/metrics.go
@@ -240,9 +240,6 @@ func serverAttrs(dst string) trace.SpanStartOption {
 		if port, err := strconv.Atoi(port); err == nil {
 			return trace.WithAttributes(attribute.String("server.address", addr), attribute.Int("server.port", port))
 		}
-
-		return trace.WithAttributes(attribute.String("server.address", addr))
 	}
-
 	return trace.WithAttributes(attribute.String("server.address", dst))
 }
--- a/valkeyotel/metrics.go
+++ b/valkeyotel/metrics.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/tls"
 	"net"
+	"strconv"
 	"sync/atomic"
 	"time"
 
@@ -182,7 +183,7 @@ func newClient(opts ...Option) (*otelclient, error) {
 
 func trackDialing(m dialMetrics, t dialTracer, dialFn func(context.Context, string, *net.Dialer, *tls.Config) (conn net.Conn, err error)) func(context.Context, string, *net.Dialer, *tls.Config) (conn net.Conn, err error) {
 	return func(ctx context.Context, dst string, dialer *net.Dialer, tlsConfig *tls.Config) (conn net.Conn, err error) {
-		ctx, span := t.Start(ctx, "valkey.dial", kind, trace.WithAttributes(dbattr, attribute.String("server.address", dst)), t.tAttrs)
+		ctx, span := t.Start(ctx, "valkey.dial", kind, trace.WithAttributes(dbattr), serverAttrs(dst), t.tAttrs)
 		defer span.End()
 
 		m.attempt.Add(ctx, 1, m.addOpts...)
@@ -232,4 +233,16 @@ func defaultDialFn(ctx context.Context, dst string, dialer *net.Dialer, cfg *tls
 		return td.DialContext(ctx, "tcp", dst)
 	}
 	return dialer.DialContext(ctx, "tcp", dst)
+}
+
+func serverAttrs(dst string) trace.SpanStartOption {
+	if addr, port, err := net.SplitHostPort(dst); err == nil {
+		if port, err := strconv.Atoi(port); err == nil {
+			return trace.WithAttributes(attribute.String("server.address", addr), attribute.Int("server.port", port))
+		}
+
+		return trace.WithAttributes(attribute.String("server.address", addr))
+	}
+
+	return trace.WithAttributes(attribute.String("server.address", dst))
 }
--- a/valkeyotel/trace_test.go
+++ b/valkeyotel/trace_test.go
@@ -376,15 +376,16 @@ func TestNewClientSimple(t *testing.T) {
 		t.Fatal("could not find SET span")
 	}
 	commandSpan := spans[commandSpanIdx]
-	validateSpanHasAttribute(t, commandSpan, "any", "label")
+	validateSpanHasAttribute(t, commandSpan, attribute.String("any", "label"))
 
 	dialSpanIdx := slices.IndexFunc(spans, func(span trace.ReadOnlySpan) bool { return span.Name() == "valkey.dial" })
 	if dialSpanIdx == -1 {
 		t.Fatal("could not find dial span")
 	}
 	dialSpan := spans[dialSpanIdx]
-	validateSpanHasAttribute(t, dialSpan, "server.address", "127.0.0.1:6379")
-	validateSpanHasAttribute(t, dialSpan, "any", "label")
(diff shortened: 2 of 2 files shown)


valkeyotel/metrics.go
```diff
@@ -182,7 +184,13 @@ func newClient(opts ...Option) (*otelclient, error) {
 
 func trackDialing(m dialMetrics, t dialTracer, dialFn func(context.Context, string, *net.Dialer, *tls.Config) (conn net.Conn, err error)) func(context.Context, string, *net.Dialer, *tls.Config) (conn net.Conn, err error) {
 	return func(ctx context.Context, dst string, dialer *net.Dialer, tlsConfig *tls.Config) (conn net.Conn, err error) {
-		ctx, span := t.Start(ctx, "valkey.dial", kind, trace.WithAttributes(dbattr, attribute.String("server.address", dst)), t.tAttrs)
+		attrs := trace.WithAttributes(dbattr, semconv.ServerAddress(dst))
+		if addr, port, err := net.SplitHostPort(dst); err == nil {
+			if port, err := strconv.Atoi(port); err == nil {
+				attrs = trace.WithAttributes(dbattr, semconv.ServerAddress(addr), semconv.ServerPort(port))
+			}
```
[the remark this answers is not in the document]
austindrenski: This is intentional to preserve the original behavior in the event that `dst` isn't what we think it is. _(Open to feedback on this if anyone feels strongly that it needs a more elaborate fallback, but for now this seemed reasonable.)_

valkeyotel/metrics.go
```diff
@@ -12,6 +13,7 @@ import (
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/metric"
+	semconv "go.opentelemetry.io/otel/semconv/v1.30.0"
```
rueian: Do we really need to import this? Pining 1.30.0 here looks weird, and it doesn't seem to be better than hard-coding the attr keys.

valkeyotel/metrics.go
```diff
@@ -182,7 +184,13 @@ func newClient(opts ...Option) (*otelclient, error) {
 
 func trackDialing(m dialMetrics, t dialTracer, dialFn func(context.Context, string, *net.Dialer, *tls.Config) (conn net.Conn, err error)) func(context.Context, string, *net.Dialer, *tls.Config) (conn net.Conn, err error) {
 	return func(ctx context.Context, dst string, dialer *net.Dialer, tlsConfig *tls.Config) (conn net.Conn, err error) {
-		ctx, span := t.Start(ctx, "valkey.dial", kind, trace.WithAttributes(dbattr, attribute.String("server.address", dst)), t.tAttrs)
+		attrs := trace.WithAttributes(dbattr, semconv.ServerAddress(dst))
```
rueian:
```suggestion
		var attrs trace.SpanStartEventOption
```

valkeyotel/metrics.go
```diff
@@ -182,7 +184,13 @@ func newClient(opts ...Option) (*otelclient, error) {
...
+		if addr, port, err := net.SplitHostPort(dst); err == nil {
+			if port, err := strconv.Atoi(port); err == nil {
+				attrs = trace.WithAttributes(dbattr, semconv.ServerAddress(addr), semconv.ServerPort(port))
+			}
+		}
```
rueian:
```suggestion
		} else {
		    attrs = trace.WithAttributes(dbattr, attribute.String("server.address", dst))
		}
```

valkeyotel/metrics.go
rueian:
```suggestion
			attrs = trace.WithAttributes(dbattr, attribute.String("server.address", addr), attribute.Int("server.port", port))
```

valkeyotel/metrics.go
austindrenski: Nope, good point. Just a habit of importing semconv in other projects, but not required here so I'll back that out this morning!

austindrenski: Nope, good point. Just a habit of importing semconv in other projects, but not required here so I'll back that out this morning!

austindrenski:
> Actually, I think with a few tweaks, this patch could cover some additional cases where valkey-go isn't sending `server.*` attrs.
> 
> I'm going to open a separate PR with that approach here in a few minutes. _(Leaving this PR as-is in case there are higher-level design decisions that I'm unaware of that might explain the lack of these `server.*` attrs in the additional cases I'm working on.)_

Opened #95 with what IMO is a better solution, or at least, a better solution for anyone creating clients with the pattern `valkeyotel.NewClient(valkey.MustParseURL("..."))`.

Leaving this PR open for contrast, but since this only fixes `server.address` for `valkey.dial` spans, users will still end up with multiple inferred entities in, e.g., the Datadog UI. Whereas, if #95 is accepted, we should see just one inferred entity.

In other words:
- before:
  ```yaml
  valkey.dial:
    db.system: "valkey"
    server.address: "{host}:{port}"
  valkey.query GET:
    db.system: "valkey"
  ```
- after #94:
  ```yaml
  valkey.dial:
    db.system: "valkey"
    server.address: "{host}"
    server.address: {port}
  valkey.query GET:
    db.system: "valkey"
  ```
- after #95:
  ```yaml
  valkey.dial:
    db.system: "valkey"
    server.address: "{host}"
    server.address: {port}
  valkey.query GET:
    db.system: "valkey"
    server.address: "{host}"
    server.address: {port}
  ```

valkeyotel/metrics.go
```diff
@@ -233,3 +234,15 @@ func defaultDialFn(ctx context.Context, dst string, dialer *net.Dialer, cfg *tls
...
+func serverAttrs(dst string) trace.SpanStartOption {
+	if addr, port, err := net.SplitHostPort(dst); err == nil {
+		if port, err := strconv.Atoi(port); err == nil {
+			return trace.WithAttributes(attribute.String("server.address", addr), attribute.Int("server.port", port))
+		}
+
+		return trace.WithAttributes(attribute.String("server.address", addr))
+	}
+
+	return trace.WithAttributes(attribute.String("server.address", dst))
```
rueian:
```suggestion
	}
	return trace.WithAttributes(attribute.String("server.address", dst))
```

PR #95, opened 2025-12-16
Set `server.*` attrs in more cases

austindrenski:
From the [OTel semantic conventions](https://opentelemetry.io/docs/specs/semconv/registry/attributes/server/) docs:

> # Server
> 
> ## Server Attributes
> 
> These attributes may be used to describe the server in a connection-based network interaction where there is one side that initiates the connection (the client is the side that initiates the connection). This covers all TCP network interactions since TCP is connection-based and one side initiates the connection (an exception is made for peer-to-peer communication over TCP where the "user-facing" surface of the protocol / API doesn't expose a clear notion of client and server). This also covers UDP network interactions where one side initiates the interaction, e.g. QUIC (HTTP/3) and DNS.
> 
> **Attributes:**
> 
> | Key | Stability | Value Type | Description | Example Values |
> | --- | --- | --- | --- | --- |
> | <a id="server-address" href="#server-address">`server.address`</a> | Stable | string | Server domain name if available without reverse DNS lookup; otherwise, IP address or Unix domain socket name. [1] | `example.com`; `10.1.2.80`; `/tmp/my.sock` |
> | <a id="server-port" href="#server-port">`server.port`</a> | Stable | int | Server port number. [2] | `80`; `8080`; `443` |
> 
> **[1] `server.address`:** When observed from the client side, and when communicating through an intermediary, `server.address` SHOULD represent the server address behind any intermediaries, for example proxies, if it's available.
> 
> **[2] `server.port`:** When observed from the client side, and when communicating through an intermediary, `server.port` SHOULD represent the server port behind any intermediaries, for example proxies, if it's available.

## Related

- https://github.com/valkey-io/valkey-go/pull/94#issuecomment-3661618076

### DIFF
--- a/valkeyotel/metrics.go
+++ b/valkeyotel/metrics.go
@@ -62,7 +62,7 @@ type dialMetrics struct {
 
 type dialTracer struct {
 	trace.Tracer
-	tAttrs trace.SpanStartEventOption
+	tAttrs []trace.SpanStartOption
 }
 
 // WithHistogramOption sets the HistogramOption.
@@ -83,6 +83,10 @@ func NewClient(clientOption valkey.ClientOption, opts ...Option) (valkey.Client,
 	oclient, err := newClient(opts...)
 	if err != nil {
 		return nil, err
+	}
+
+	if len(clientOption.InitAddress) == 1 {
+		oclient.tAttrs = append(oclient.tAttrs, serverAttrs(clientOption.InitAddress[0]))
 	}
 
 	if clientOption.DialCtxFn == nil {
@@ -135,9 +139,7 @@ func NewClient(clientOption valkey.ClientOption, opts ...Option) (valkey.Client,
 }
 
 func newClient(opts ...Option) (*otelclient, error) {
-	cli := &otelclient{
-		tAttrs: trace.WithAttributes(),
-	}
+	cli := &otelclient{}
 	for _, opt := range opts {
 		opt(cli)
 	}
@@ -183,7 +185,7 @@ func newClient(opts ...Option) (*otelclient, error) {
 
 func trackDialing(m dialMetrics, t dialTracer, dialFn func(context.Context, string, *net.Dialer, *tls.Config) (conn net.Conn, err error)) func(context.Context, string, *net.Dialer, *tls.Config) (conn net.Conn, err error) {
 	return func(ctx context.Context, dst string, dialer *net.Dialer, tlsConfig *tls.Config) (conn net.Conn, err error) {
-		ctx, span := t.Start(ctx, "valkey.dial", kind, trace.WithAttributes(dbattr), serverAttrs(dst), t.tAttrs)
+		ctx, span := t.Start(ctx, "valkey.dial", append(t.tAttrs, kind, trace.WithAttributes(dbattr), serverAttrs(dst))...)
 		defer span.End()
 
 		m.attempt.Add(ctx, 1, m.addOpts...)
--- a/valkeyotel/metrics.go
+++ b/valkeyotel/metrics.go
@@ -62,7 +62,7 @@ type dialMetrics struct {
 
 type dialTracer struct {
 	trace.Tracer
-	tAttrs []trace.SpanStartOption
+	tAttrs trace.SpanStartEventOption
 }
 
 // WithHistogramOption sets the HistogramOption.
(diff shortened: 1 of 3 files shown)


valkeyotel/metrics.go
```diff
@@ -84,6 +85,10 @@ func NewClient(clientOption valkey.ClientOption, opts ...Option) (valkey.Client,
 		return nil, err
 	}
 
+	if len(clientOption.InitAddress) == 1 {
+		oclient.tAttrs = append(oclient.tAttrs, serverAttrs(clientOption.InitAddress[0]))
+	}
```
austindrenski, line 90: This obviously won't cover everything, but it should cover some very common cases (e.g. AWS ElastiCache serverless clusters) and leave room for others with more complex topologies to contribute further detections in the future.

valkeyotel/trace.go
```diff
@@ -163,7 +165,7 @@ type otelclient struct {
 	meter           metric.Meter
 	cscMiss         metric.Int64Counter
 	cscHits         metric.Int64Counter
-	tAttrs          trace.SpanStartEventOption
+	tAttrs          []trace.SpanStartOption
```
austindrenski: _tbh this should probably be renamed to `spanOpts`, but renaming feels out of scope for this PR, so leaving it for a follow up if/when this PR is accepted._

valkeyotel/trace.go
```diff
@@ -504,8 +506,8 @@ func (d *dedicated) end(span trace.Span, err error) {
 	endSpan(span, err)
 }
 
-func startSpan(tracer trace.Tracer, ctx context.Context, op string, size int, attrs trace.SpanStartEventOption) (context.Context, trace.Span) {
-	return tracer.Start(ctx, op, kind, attr(op, size), attrs)
+func startSpan(tracer trace.Tracer, ctx context.Context, op string, size int, attrs []trace.SpanStartOption) (context.Context, trace.Span) {
+	return tracer.Start(ctx, op, append(attrs, kind, attr(op, size))...)
```
austindrenski: There are a couple of places where I've changed `SpanStartEventOption` to `SpanStartOption` so that when we can play nice with `kind` when we get down into here.

valkeyotel/metrics.go
```diff
@@ -182,7 +185,7 @@ func newClient(opts ...Option) (*otelclient, error) {
 
 func trackDialing(m dialMetrics, t dialTracer, dialFn func(context.Context, string, *net.Dialer, *tls.Config) (conn net.Conn, err error)) func(context.Context, string, *net.Dialer, *tls.Config) (conn net.Conn, err error) {
 	return func(ctx context.Context, dst string, dialer *net.Dialer, tlsConfig *tls.Config) (conn net.Conn, err error) {
-		ctx, span := t.Start(ctx, "valkey.dial", kind, trace.WithAttributes(dbattr, attribute.String("server.address", dst)), t.tAttrs)
+		ctx, span := t.Start(ctx, "valkey.dial", append(t.tAttrs, kind, trace.WithAttributes(dbattr), serverAttrs(dst))...)
```
austindrenski: There's a small behavioral change here, where previously if someone was explicitly setting `server.{address,port}` by calling `otelvalkey.TraceAttrs(...)` it would have _overridden_ the erroneous `server.address` that we were setting previously.

Whereas now we will overwrite their attempted override, but since part of this PR is fixing the `server.address` attrs, I _personally_ wouldn't consider this a "breaking" change, but others may disagree (e.g. maybe someone out there is trying to intentionally vanity-plate some internal IPs, etc).

It seems unlikely that anyone who was doing this^ would be _harmed_ by this reordering, but again, anything is possible, so open to hearing what others have to say on it.

valkeyotel/trace.go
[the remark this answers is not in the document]
austindrenski: bad bot

austindrenski:
> I think these tests are failing on account of them using the deprecated `WithClient(...)` method instead of the `NewClient(...)` method.
> 
> Will take a look and see what is happening this morning.

Turns out that the `validateTrace(...)` helper was explicitly indexing span attrs instead of using `validateSpanHasAttribute(...)`, so updated `validateTrace(...)` and should be good to review now @rueian.

valkeyotel/metrics.go
rueian: I don't consider that a breaking change either. But can we keep `tAttrs` a `trace.SpanStartEventOption`? I am afraid that doing `append()...` is costly.

austindrenski: Hmmm, actually, I think we can if I just add another `*Attrs` to the structs (i.e. so we're still passing individual values into the `...trace.SpanStartOptions` instead).

austindrenski: Alright, updated to remove all of those appends. Ready for another look @rueian!

rueian (approved):

PR #102, opened 2025-12-28
fix: valkeycompat NewAdapter sets maxp incorrectly

### LINKED ISSUE #96: valkeycompat NewAdapter sets maxp incorrectly
If I have misunderstood this then my apologies.

The valkeycompat.NewAdapter sets maxp to the current no. of cores. This is then used to scale out commands.
Even if one is running on 1 or less cores most operations are IO bound so this limit is unnecessarily restrictive.

Secondly this kind of setting should be user-controlled and not something arbitrarily set internally.

A better interface would be to add an option that allows the user to set this themselves using the functional option pattern.

majiayu000:
## Summary
This PR fixes #96

## Changes
- `valkeycompat/adapter.go`
- `valkeycompat/adapter_option_test.go`

### DIFF
--- a/valkeycompat/adapter.go
+++ b/valkeycompat/adapter.go
@@ -639,8 +639,23 @@ type CacheCompat struct {
 	ttl    time.Duration
 }
 
-func NewAdapter(client valkey.Client) Cmdable {
-	return &Compat{client: client, maxp: runtime.GOMAXPROCS(0)}
+// AdapterOption is a functional option type for NewAdapter
+type AdapterOption func(c *Compat)
+
+// WithNodeScaleoutLimit sets the maximum parallelism for node scaleout operations.
+// If not set, defaults to runtime.GOMAXPROCS(0).
+func WithNodeScaleoutLimit(limit int) AdapterOption {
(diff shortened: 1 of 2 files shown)


valkeycompat/adapter.go
```diff
@@ -639,8 +639,23 @@ type CacheCompat struct {
...
-func NewAdapter(client valkey.Client) Cmdable {
-	return &Compat{client: client, maxp: runtime.GOMAXPROCS(0)}
+// AdapterOption is a functional option type for NewAdapter
+type AdapterOption func(c *Compat)
+
+// WithNodeScaleoutLimit sets the maximum parallelism for node scaleout operations.
+// If not set, defaults to runtime.GOMAXPROCS(0).
+func WithNodeScaleoutLimit(limit int) AdapterOption {
+	return func(c *Compat) {
+		c.maxp = limit
```
rueian, line 653: We many need to avoid it from being set to less than 1.

majiayu000:
Fixed. Values less than 1 will now default to 1.

rueian (approved):
Thanks!

PR #108, opened 2026-02-01
fix: race condition in Lua script SHA-1 loading

wckd:
Add double-check pattern inside the singleflight fn() to prevent duplicate SCRIPT LOAD calls. When the mock/server responds instantly, the sequence could be:
1. G1 and G2 both read s.sha1 == ""
2. G1 executes singleflight, completes, sets s.sha1, clears c.ch
3. G2 calls singleflight.Do(), sees c.ch == nil, executes fn() again

The fix re-checks s.sha1 inside the singleflight to handle this race.

### DIFF
--- a/lua.go
+++ b/lua.go
@@ -86,7 +86,6 @@ func newLuaScript(script string, readonly bool, noSha1, retryable bool, opts ...
 type Lua struct {
 	script    string
 	sha1      string
-	sha1Call  call
 	maxp      int
 	sha1Mu    sync.RWMutex
 	readonly  bool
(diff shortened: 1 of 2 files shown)


lua.go
```diff
@@ -114,6 +114,15 @@ func (s *Lua) Exec(ctx context.Context, c Client, keys, args []string) (resp Val
 		// If not loaded yet, use singleflight to load it.
 		if scriptSha1 == "" {
 			err := s.sha1Call.Do(ctx, func() error {
+				// Double-check pattern: another goroutine may have loaded SHA-1
```
rueian: Hi @wckd, thanks for your PR.

Yes, we do need the double check, and I just also realized that `sha1Call` is unnecessary. Would you mind helping us remove it? We should just do this:

```go
		// Check if SHA-1 is already loaded.
		s.sha1Mu.RLock()
		scriptSha1 = s.sha1
		s.sha1Mu.RUnlock()

		if scriptSha1 == "" {
			s.sha1Mu.Lock()
			if scriptSha1 == "" { // the double check
				result := c.Do(ctx, c.B().ScriptLoad().Script(s.script).Build().ToRetryable())
			    ...
			s.sha1Mu.Unlock()
...
```

wckd: Updated the code with your recommendation, removing the sha1Call. Please review :-)

wckd:
Added test coverage for the SCRIPT LOAD error path as flagged by codecov.

rueian (approved):

PR #110, opened 2026-02-15
feat(otel): Optimize metrics recording

gaiaz-iusipov:


valkeyotel/trace.go
```diff
@@ -115,44 +117,95 @@ type commandMetrics struct {
...
+		if opOpt != nil {
+			*opts = append(*opts, opOpt)
+		}
+		if labelerOpt != nil {
+			*opts = append(*opts, labelerOpt)
+		}
 	}
 
-	c.duration.Record(ctx, time.Since(now).Seconds(), opts...)
+	// Use floating point division here for higher precision (instead of Seconds method).
```
rueian: Could you give some examples?

gaiaz-iusipov: I thought this change made sense, similar to how it’s implemented in [opentelemetry-go-contrib](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/google.golang.org/grpc/otelgrpc/stats_handler.go#L358).

gaiaz-iusipov: At the moment it turns out that [both variants are being used](https://github.com/valkey-io/valkey-go/blob/main/valkeyotel/metrics.go#L207). I suggest using a single unified variant.

rueian (approved):

valkeyotel/trace.go
gaiaz-iusipov: @rueian what do you think about this?

Issue #111, opened 2026-02-18
valkeycompat JSONGet is incompatible with go-redis


piyongcai-liucai:
When the specified key in JSONGet does not exist, 
1. go-redis returns redis.Nil error,
2. valkey-go returns nil error.

1. go-redis
```go
ctx := context.Background()
client := redis.NewUniversalClient(&redis.UniversalOptions{
	Addrs:    []string{"localhost:6379"},
	Protocol: 3,
	PoolSize: 10,
})

// go-redis will return redis.Nil
rawOne, err := client.JSONGet(ctx, "cdcypw:stage:311118").Result()
if err != nil {
	t.Fatal(err)
}
fmt.Println(rawOne)
```

2. valkeycomnpay
```go
client, err := valkey.NewClient(valkey.ClientOption{InitAddress: []string{"127.0.0.1:6379"}})
if err != nil {
	t.Error(err)
}
cmdable := valkeycompat.NewAdapter(client)

// valkeycompay return nil error
rawOne, err := cmdable.JSONGet(t.Context(), "cdcypw:stage:311118").Result()
if err != nil {
	t.Fatal(err)
}
fmt.Println(rawOne)
```

rueian:
Hi @piyongcai-liucai, thanks for reporting. Is return redis.Nil the latest go-redis behavior?

piyongcai-liucai:
valkey-go v1.0.72 already fixed this bug.

PR #114, opened 2026-02-20
fixed #112 #111

### LINKED ISSUE #112: The functions TxPipeline, TxPipelined, Pipeline, Pipelined, valkeycompay are incompatible with Go-Redis.
1. In go-redis, if any command contains an error, it will return the error.
```go
ctx := context.Background()
client := redis.NewUniversalClient(&redis.UniversalOptions{
	Addrs:    []string{"localhost:6379"},
	Protocol: 3,
	PoolSize: 10,
})

// go-redis will return redis.Nil, but valkey return nil error
_, err = client.TxPipelined(ctx, func(piper redis.Pipeliner) error {
       // eg. JSONSet, when key not exists
       piper.JSONSet(ctx, "cdcypw:stage:311118", "status", false)  // <-- HERE
      // other command 
       return nil
})
if err != nil {
	// go-redis will return  "ERR new objects must be created at the root"  // <--- HERE
}
```

2. In valkeycompat, the command failed and the external error message remains null.
```go

client, err = valkey.NewClient(valkey.ClientOption{InitAddress: []string{"127.0.0.1:6379"}})
if err != nil {
	panic(err)
}

cmdable := valkeycompat.NewAdapter(defaultClient)

ctx := context.Background()
// go-redis will return redis.Nil, but valkey return nil error
_, err = client.TxPipelined(ctx, func(piper valkeycompat.Pipeliner) error {
       // eg. JSONSet, when key not exists
       piper.JSONSet(ctx, "cdcypw:stage:311118", "status", false)  // <-- HERE
        // other command 
       return nil
})
if err != nil {
	// valkeycompat will return  nil   // <--- HERE
}
```

piyongcai-liucai:
TxPipeline, TxPipelined, Pipeline, Pipelined compatible with go-redis


valkeycompat/pipeline.go
```diff
@@ -3207,11 +3208,17 @@ func (c *Pipeline) Exec(ctx context.Context) ([]Cmder, error) {
...
 			err = r.NonValkeyError()
 		}
 		rets[i].SetErr(nil)
 		rets[i].from(r)
+
+		if err == nil {
+			if subErr := rets[i].Err(); subErr != nil {
+				err = subErr
+			}
+		}
```
rueian: Hi @piyongcai-liucai, thanks for the PR but this doesn't look right. Maybe you could start with a test first?

PR #120, opened 2026-03-05
Unify CacheCompat to use pointer receivers.

### LINKED ISSUE #118: Unified CacheCompat method implementation
In the **CacheCompat** type, some methods are declared with a value receiver **func (c CacheCompat)**, while others use a pointer receiver **func (c *CacheCompat)**.

**Suggestion**: To unify the declaration form of these two types of methods, we recommend using the **pointer receiver** style.

The advantage is that if you only need to perform a single operation, you can complete it in **one line of code** with a pointer receiver.

If you mix value and pointer receivers and happen to call a pointer method, you will need **two lines of code** to finish the operation.

eg
```golang
client := ...
ctx :=context.Background()
key := "_some_key_"
cmdable := valkeycompat.NewAdapter(client)

// HERE, you must write two line code
cacheCompat := cmdable.Cache(time.Minute)
jsonStr, err := cacheCompat.JSONGet(sctx , key, "check_code").Result()

// if you write together, will got an error: 
// cannot call pointer method JSONGet on valkeycompat.CacheCompat
jsonStr, err = cmdable.Cache(time.Minute).JSONGet(ctx, key, "check_code").Result()

...
```

piyongcai-liucai:
fixed #118.

### DIFF
--- a/valkeycompat/adapter.go
+++ b/valkeycompat/adapter.go
@@ -264,6 +264,7 @@ type CoreCmdable interface {
 	XInfoStream(ctx context.Context, key string) *XInfoStreamCmd
 	XInfoStreamFull(ctx context.Context, key string, count int64) *XInfoStreamFullCmd
 	XInfoConsumers(ctx context.Context, key string, group string) *XInfoConsumersCmd
+	XCfgSet(ctx context.Context, a XCfgSetArgs) *StatusCmd
 
 	BZPopMax(ctx context.Context, timeout time.Duration, keys ...string) *ZWithKeyCmd
 	BZPopMin(ctx context.Context, timeout time.Duration, keys ...string) *ZWithKeyCmd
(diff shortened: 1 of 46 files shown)


valkeycompat/adapter.go
```diff
@@ -53,7 +53,7 @@ var Nil = valkey.Nil
 
 type Cmdable interface {
 	CoreCmdable
-	Cache(ttl time.Duration) CacheCompat
+	Cache(ttl time.Duration) *CacheCompat
```
rueian:
```suggestion
	Cache(ttl time.Duration) CacheCompat
```

This one is intentional to avoid heap allocation. We can't change this.

valkeycompat/adapter.go
```diff
@@ -669,8 +669,8 @@ func (c *Compat) Client() valkey.Client {
 	return c.client
 }
 
-func (c *Compat) Cache(ttl time.Duration) CacheCompat {
-	return CacheCompat{client: c.client, ttl: ttl}
+func (c *Compat) Cache(ttl time.Duration) *CacheCompat {
+	return &CacheCompat{client: c.client, ttl: ttl}
```
rueian:
```suggestion
func (c *Compat) Cache(ttl time.Duration) CacheCompat {
	return CacheCompat{client: c.client, ttl: ttl}
```

rueian (approved):

PR #123, opened 2026-03-11
feat(otel): add `WithOpNameResolver` and `WithSpanNameFormatter`

gaiaz-iusipov:
Closes: #122

### DIFF
--- a/valkeyotel/metrics.go
+++ b/valkeyotel/metrics.go
@@ -8,12 +8,13 @@ import (
 	"sync/atomic"
 	"time"
 
-	"github.com/valkey-io/valkey-go"
 	"go.opentelemetry.io/otel"
 	"go.opentelemetry.io/otel/attribute"
 	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/metric"
 	"go.opentelemetry.io/otel/trace"
+
+	"github.com/valkey-io/valkey-go"
 )
 
 var (
@@ -155,6 +156,12 @@ func newClient(opts ...Option) (*otelclient, error) {
 	if cli.tracerProvider == nil {
 		cli.tracerProvider = otel.GetTracerProvider() // Default to global TracerProvider
 	}
+	if cli.opNameResolver == nil {
+		cli.opNameResolver = &DefaultOpNameResolver{Limit: 5}
+	}
+	if cli.spanNameFormatter == nil {
+		cli.spanNameFormatter = defaultSpanNameFormatter
+	}
 
 	// Now that we have the meterProvider and tracerProvider, get the Meter and Tracer
 	cli.meter = cli.meterProvider.Meter(name)
@@ -245,4 +252,8 @@ func serverAttrs(dst string) trace.SpanStartEventOption {
 		}
 	}
 	return trace.WithAttributes(attribute.String("server.address", dst))
+}
+
+func defaultSpanNameFormatter(_ context.Context, op string) string {
+	return op
 }
--- a/valkeyotel/op_name.go
+++ b/valkeyotel/op_name.go
@@ -1,0 +1,78 @@
+package valkeyotel
+
+import (
+	"context"
+	"strings"
+
+	"github.com/valkey-io/valkey-go"
+)
+
+type OpNameResolver interface {
+	OpName(ctx context.Context, cmd valkey.Completed) string
+	MultiOpName(ctx context.Context, cmds valkey.Commands) string
+	MultiCacheableOpName(ctx context.Context, cmds []valkey.CacheableTTL) string
+}
+
+var _ OpNameResolver = (*DefaultOpNameResolver)(nil)
+
+type DefaultOpNameResolver struct {
+	// Limit controls how many elements are used to compose the resulting operation name.
+	// If Limit is greater than zero, only the first Limit commands are used.
+	Limit int
+}
+
+func (DefaultOpNameResolver) OpName(_ context.Context, cmd valkey.Completed) string {
+	return cmd.Commands()[0]
+}
+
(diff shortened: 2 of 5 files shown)


valkeyotel/op_name.go
```diff
@@ -0,0 +1,78 @@
+package valkeyotel
+
+import (
+	"context"
+	"strings"
+
+	"github.com/valkey-io/valkey-go"
+)
+
+type OpNameResolver interface {
```
rueian, line 10:
```suggestion
type OpNameFormater interface {
```

valkeyotel/op_name.go
```diff
@@ -0,0 +1,78 @@
...
+	if r.Limit > 0 && len(cmds) > r.Limit {
+		cmds = cmds[:r.Limit]
+	}
+
+	size := len(cmds) - 1
+	for _, cmd := range cmds {
+		size += len(cmd.Commands()[0])
+	}
+
+	sb := new(strings.Builder)
```
rueian: will this be allocated on the heap?

valkeyotel/op_name.go
```diff
@@ -0,0 +1,78 @@
...
+	sb.Grow(size)
+	sb.WriteString(cmds[0].Commands()[0])
+	for _, cmd := range cmds[1:] {
```
rueian, line 47: why to exclude the first cmd?

valkeyotel/op_name.go
```diff
@@ -0,0 +1,78 @@
...
+
+	size := len(cmds) - 1
+	for _, cmd := range cmds {
+		size += len(cmd.Cmd.Commands()[0])
+	}
+
+	sb := new(strings.Builder)
+	sb.Grow(size)
+	sb.WriteString(cmds[0].Cmd.Commands()[0])
+	for _, cmd := range cmds[1:] {
```
rueian, line 73: why to exclude the first cmd?

valkeyotel/metrics.go
```diff
@@ -246,3 +253,7 @@ func serverAttrs(dst string) trace.SpanStartEventOption {
 	}
 	return trace.WithAttributes(attribute.String("server.address", dst))
 }
+
+func defaultSpanNameFormatter(_ context.Context, op string) string {
+	return op
```
rueian, line 258: Hmm, how can a user use this? If the input is a string already, the user can't do much with it.

valkeyotel/op_name.go
gaiaz-iusipov: At this point in the code, we know for sure that the slice contains at least two values. Therefore, to avoid `if` checks on each loop iteration when writing the separator (a space character), we first write the first element before the loop and then write all subsequent elements inside the loop.

valkeyotel/op_name.go
gaiaz-iusipov: At this point in the code, we know for sure that the slice contains at least two values. Therefore, to avoid `if` checks on each loop iteration when writing the separator (a space character), we first write the first element before the loop and then write all subsequent elements inside the loop.

valkeyotel/metrics.go
gaiaz-iusipov: `spanNameFormatter` is implemented similarly to how it is done in the instrumentation from `open-telemetry`: essentially, we already have a `spanName`, and it is assumed (but not limited to) that the client may want to format it. However, since the context is available, the client remains free to use any strategy for choosing the span name.

valkeyotel/op_name.go
gaiaz-iusipov: The name `OpNameFormatter` implies that some name already exists and the function formats it. In this case the name does not exist yet, it is determined (created) based on the command and the context.

valkeyotel/op_name.go
gaiaz-iusipov: Fixed

valkeyotel/op_name.go
rueian: nice

valkeyotel/metrics.go
rueian: Are you saying the othttp plugin? I took a look at https://github.com/open-telemetry/opentelemetry-go/pull/617/changes and found their spanNameFormatter signature was richer: `func(operation string, r *http.Request) string`.

But, if we have OpNameResolver already, do we still need to provide spanNameFormatter?

gaiaz-iusipov: Yes, the idea is that if a client has some standardized notation for span names or wants to extend the operation name (for example, with a suffix containing an A/B test group, a server identifier if there are several, etc.). At the same time, accepting the command object as an argument seemed excessive to me and would require introducing an interface.

rueian: Isn’t opNameResolver alone enough?

gaiaz-iusipov: It is desirable to leave the choice to the client between the metric attribute (low cardinality, no unnecessary details, changes require updates to dashboards) and the span name (no cardinality restrictions, focus on clarity, and it is acceptable to include details in the name). Moreover, this capability is practically free in terms of resource usage.

gaiaz-iusipov: In the case of the metric, there is already some Valkey context (the metric name), so it is clear what the operation name refers to. In the case of spans, it might end up being just `"GET"`, and it would be unclear what span it is. Let's change the span names to `"Valkey ${opName}"`, [similar to how it is done in the HTTP client](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/instrumentation/net/http/otelhttp/transport.go#L81). I will remove `spanNameFormatter` from the PR.

rueian: no need. let's keep it.

rueian (approved):

Issue #75, opened 2025-09-02
Feature Request: Advanced Zone-Aware Read Routing


PingXie:
### Summary

Currently, valkey-go supports basic zone-aware routing by allowing reads to be directed to replicas. However, this mechanism lacks the context to implement more sophisticated and cost-effective routing strategies where nodes are not always evenly distributed across Availability Zones.

This proposal introduces a new, optional callback, `ReadFromNodeSelector`, to provide the client with the full cluster topology context for a given keyslot, enabling intelligent, application-defined read routing.

The Problem: Inefficient Routing with Uneven Node Distribution

  The core challenge arises when the distribution of clients across AZs does not match the distribution of Valkey nodes (primaries and replicas)
  for a given slot. The current `SendToReplicas` and `ReplicaSelector` callbacks do not provide enough information to handle these scenarios
  optimally.

  Consider a common, cost-conscious deployment:
   * 3 Availability Zones: us-east-1a, us-east-1b, us-east-1c
   * 1 Primary + 1 Replica: For a given slot, the primary is in us-east-1b and the replica is in us-east-1c.
   * No node for this slot exists in `us-east-1a`.

  Now, let's analyze the behavior of a client based on its location:

   1. Client in `us-east-1a`: This client has no local node for the data. The ideal behavior is to treat the primary (1b) and the replica (1c) as
      co-equals for reads, spreading the load evenly between them to maximize throughput and availability. The current system makes it difficult to implement this "fallback" logic, as the primary is not typically included in the replica selection process.

   2. Client in `us-east-1b`: This client is in the same AZ as the primary. To minimize latency and avoid cross-AZ data transfer costs, 100% of its
      reads should go to the local primary node. It should never send a read to the replica in us-east-1c. The current replica-only selection logic
      might inadvertently route traffic across zones, incurring unnecessary costs.

  The fundamental principle is: always prefer a node in the local zone if one exists, regardless of whether it's a primary or a replica. Only if
  no local node exists should the client distribute reads across all available nodes in other zones.

###  Proposed Solution: `ReadFromNodeSelector`

```golang
    1 type NodeInfo struct {
    2     Addr string
    3     AZ   string
    4 }
    5
    6 type ClientOption struct {
    7     // ... existing fields
    8     AvailabilityZone string
    9     ReadFromNodeSelector func(cmd Completed, clientAZ string, nodes []NodeInfo) string
   10 }
```
  How it works:

[...]

PR #125, opened 2026-04-05
Make ReadNodeSelector return only replicas

Luis729:
Right now we use logic that enables `SendtoReplicas`, which lets you also customize `ReadNodeSelector`. The weird thing is that even though `SendToReplicas` is true, `ReadNodeSelector` still lets you choose a primary.

Just from the understanding. We have two cases:

1. SendToReplicas = false: ReadNodeSelector never triggered, and command sent to the primary node
2. SendToReplicas = true: ReadNodeSelector triggered, and command set to the replica nodes

But the existing code does not do (2). This PR aims to do that to align on what the variable states.

----

**IMPORTANT:** This is a breaking change for people who expected a primary from ReadNodeSelector btw. But they can still get the primary by passing "-1" instead of 0. Or passing a number out of index (lets say 999999)

And to be frank, this should be the right behavior. ReadNodeSelector, given its exclusive to SendToReplicas, should only read replica indexes. So replica 1 should be index 0 and so forth...

Before, lets say you wanted to have a random node read:

```go
// before
cfg.ReadNodeSelector = func(slot uint16, nodes []valkey.NodeInfo) int {
    if len(nodes) <= 1 {
        return 0
    }
    return rand.IntN(len(nodes)-1) + 1 // skip index 0 (assumed primary)
}

// after
cfg.ReadNodeSelector = func(slot uint16, nodes []valkey.NodeInfo) int {
    if len(nodes) == 0 {
        return -1
    }
    return rand.IntN(len(nodes))
}
```

---

https://github.com/redis/rueidis/discussions/688#discussioncomment-11510498

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -463,9 +463,11 @@ func (c *clusterClient) _pick(slot uint16, toReplica bool) (p conn) {
 		if nodes := c.rslots[slot]; len(nodes) > 0 {
 			rIndex := 0
 			if c.opt.ReadNodeSelector != nil {
-				rIndex = c.opt.ReadNodeSelector(slot, nodes)
-				if rIndex < 0 || rIndex >= len(nodes) {
+				rIndex = c.opt.ReadNodeSelector(slot, nodes[1:])
+				if rIndex < 0 || rIndex >= len(nodes)-1 {
 					rIndex = 0
+				} else {
+					rIndex = rIndex + 1
 				}
 			}
 			p = nodes[rIndex].conn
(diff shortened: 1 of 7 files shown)


rueian:
Hi @Luis729, please take a look at https://github.com/valkey-io/valkey-go/issues/75 for why `ReadNodeSelector` contains the primary.

If you don't want that you can just use ReplicaSelector instead.

Luis729:
@rueian Thanks for that, I was wondering why `ReplicaSelector` is deprecated if that's the case. I noticed `ReadNodeSelector` replaces it, and if it's for AZ cases, then I understand. But if we want ReadNodeSelector to be "read only nodes" then shouldn't that be replicas only? I am curious to know about the take here.

rueian:
> But if we want ReadNodeSelector to be "read only nodes" then shouldn't that be replicas only?

It is just to keep the input of ReadNodeSelector consistent. If the primary is part of the input, then we always include it.

`SendToReplicas` is not a good name now. Please treat it as "SendToReadNode".

Luis729:
Thanks for the reply @rueian, will close this PR out.

PR #128, opened 2026-04-18
Call hmset from valkeycompat hmset

Luis729:
We want valkeycompat's HMSet command to behave the same as go-redis. Not sure if that was a typo but setting it to HMSet (even if its deprecated).

```go
goredisClient.Del(ctx, "go:h")
valkeyCompatClient.Del(ctx, "vk:h")

goredisClient.HMSet(ctx, "go:h", "f1", "v1", "f2", "v2")      // .Val() = true
valkeyCompatClient.HMSet(ctx, "vk:h", "f1", "v1", "f2", "v2") // .Val() = true

goredisClient.HMSet(ctx, "go:h", "f1", "v1", "f2", "v2")      // .Val() = true
valkeyCompatClient.HMSet(ctx, "vk:h", "f1", "v1", "f2", "v2") // .Val() = false
```


rueian (approved):
Yes it was because HMSET is deprecated

PR #130, opened 2026-04-30
valkeycompat: Make StringCmd.Bool() match go-redis

Luis729:
(we were migrating some client and noticed this, figure to add it upstream too, thanks!)

Reference (go-redis v9.19.0):
https://github.com/redis/go-redis/blob/v9.19.0/command.go#L1383-L1388


rueian (approved):
Thank you @Luis729!

PR #132, opened 2026-05-06
Pin dependencies in GitHub Workflows and dockertest.sh

davquar:
Hi, I noticed from the OpenSSF scorecard for this project that some dependencies are not pinned: https://securityscorecards.dev/viewer/?uri=github.com/valkey-io/valkey-go.

As pinning dependencies is generally a good security practice to reduce the risk of supply-chain attacks, I've made the following changes:

- In `.github/workflows`: pin all GitHub Actions dependencies by hash, with a comment helping out to identify the version tag.
- In `dockertest.sh`, replace `latest` with the version tag, which is sufficient given the gosumdb infra.

With these changes, the OpenSSF score for `Pinned-Dependencies` goes from `1/10` to `10/10`:

### Before

```
❯ scorecard --checks=Pinned-Dependencies --repo=valkey-io/valkey-go
Starting (valkey-io/valkey-go) [Pinned-Dependencies]
Finished (valkey-io/valkey-go) [Pinned-Dependencies]

RESULTS
-------
Aggregate score: 1.0 / 10

Check scores:
|--------|---------------------|-----------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------|
| SCORE  |        NAME         |                             REASON                              |                                            DOCUMENTATION / REMEDIATION                                             |
|--------|---------------------|-----------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------|
| 1 / 10 | Pinned-Dependencies | dependency not pinned by hash detected -- score normalized to 1 | https://github.com/ossf/scorecard/blob/c395761df6afe1a69e476bc60a013a94bcbc153f/docs/checks.md#pinned-dependencies |
|--------|---------------------|-----------------------------------------------------------------|--------------------------------------------------------------------------------------------------------------------|
```

### After

```
❯ scorecard --checks=Pinned-Dependencies --local .
Starting (.) [Pinned-Dependencies]
Finished (.) [Pinned-Dependencies]

RESULTS
-------
Aggregate score: 10.0 / 10

Check scores:
|---------|---------------------|-----------------------------|--------------------------------------------------------------------------------------------------------------------|
|  SCORE  |        NAME         |           REASON            |                                            DOCUMENTATION / REMEDIATION                                             |
|---------|---------------------|-----------------------------|--------------------------------------------------------------------------------------------------------------------|
| 10 / 10 | Pinned-Dependencies | all dependencies are pinned | https://github.com/ossf/scorecard/blob/c395761df6afe1a69e476bc60a013a94bcbc153f/docs/checks.md#pinned-dependencies |
|---------|---------------------|-----------------------------|--------------------------------------------------------------------------------------------------------------------|
```

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -18,7 +18,7 @@ jobs:
   validate-modules:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
       - name: Validate Required Modules
         run: |
           # Find all module directories: valkey* (except valkeylock), om, and mock
@@ -39,7 +39,7 @@ jobs:
     outputs:
       matrix: ${{ steps.set-matrix.outputs.matrix }}
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
       - id: set-matrix
         run: |
           echo "matrix=$(find . -maxdepth 2 -type f -name 'go.mod' | xargs -n 1 dirname | sort -u | { echo "e2e"; cat; } | jq -R -s -c 'split("\n")[:-1]')" >> $GITHUB_OUTPUT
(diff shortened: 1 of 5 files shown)


EvgeniyPatlan (approved):

.github/workflows/build.yml
```diff
@@ -18,7 +18,7 @@ jobs:
   validate-modules:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
```
rueian, line 21: Can’t it be just 4.3.1?

davquar: Unfortunately won't help, as tags in GitHub Actions are not immutable. From GitHub's [secure use reference](https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions):

> **Pin actions to a full-length commit SHA**
> Pinning an action to a full-length commit SHA is currently the only way to use an action as an immutable release. Pinning to a particular SHA helps mitigate the risk of a bad actor adding a backdoor to the action's repository, as they would need to generate a SHA-1 collision for a valid Git object payload. When selecting a SHA, you should verify it is from the action's repository and not a repository fork.

rueian (approved):

PR #137, opened 2026-05-14
Add remaining commands to valkeycompatmock

Luis729:
Adds remaining valkeycompatmock commands so the Expect surface covers go-redis/redismock parity.

317/320 redismock Expect* are mirrored.
- Missing: ExpectDo, ExpectFCallRo, ExpectWatch (no equivalent in valkeycompat so will omit these here too)

To make sure they work 1-1 with redismock: https://github.com/Luis729/valkey-go/tree/paul-valkeycompat-mock-conformance/conformance/valkeycompatmock


Luis729:
@rueian this PR is massive - I'm not sure who else can review this, but I figure I added the remainder of the commands

valkeycompatmock/go.mod
```diff
@@ -9,10 +9,16 @@ replace github.com/valkey-io/valkey-go/mock => ../mock
 replace github.com/valkey-io/valkey-go/valkeycompat => ../valkeycompat
 
 require (
+	github.com/go-redis/redismock/v9 v9.2.0
+	github.com/redis/go-redis/v9 v9.2.0
```
rueian: Is it possible that we don't import these?

Luis729: Done, yeah it was for my conformance testing, which I will update the PR description with to basically get this package's behavior as close to go-redis/redismock as possible

rueian (approved):

PR #139, opened 2026-05-20
feat: impl WithOnReceiveReturnHook

### LINKED ISSUE #134: SUBSCRIBE doesn't get unsubscribed
I was trying to use `Client.Receive()` for a temporary PUBSUB, but noticed that valkey is leaving subscriptions active (confirmed using `PUBSUB CHANNELS`) even after the context that's passed into Receive() is cancelled - this seems counterintuitive, as Receive() doesn't return an object that could be used for closing the subscription, therefore cancelling the context seems the best way to indicate "stop receiving on this channel".

I'm aware that the docs also mention calling `UNSUBSCRIBE` manually, but how would the library know which subscription am I trying to unsubscribe from? (given there could be _n_ calls to Receive on the same channel in progress)

rueian:
Closes https://github.com/valkey-io/valkey-go/issues/134

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,7 +332,7 @@ err := client.Receive(ctx, client.B().Subscribe().Channel("news").Build(), func(
 
 #### Cleanup on Receive return
 
-Note that `client.Receive()` doesn't unsubscribe channels automatically. Use `valkey.WithOnReceiveReturnHook` when you need to execute cleanup commands (e.g., `UNSUBSCRIBE`) on the underlying connection when a `client.Receive()` is about to return.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -330,6 +330,27 @@ err := client.Receive(ctx, client.B().Subscribe().Channel("news").Build(), func(
 })
 ```
 
+#### Cleanup on Receive return
+
+Note that `client.Receive()` doesn't unsubscribe channels automatically. Use `valkey.WithOnReceiveReturnHook` when you need to execute cleanup commands (e.g., `UNSUBSCRIBE`) on the underlying connection when a `client.Receive()` is about to return.
```
rueian:
```suggestion
Note that `client.Receive()` doesn't unsubscribe channels automatically when it is about to return. Use `valkey.WithOnReceiveReturnHook` if you need to execute cleanup commands (e.g., `UNSUBSCRIBE`) on the underlying connection.
```

Issue #140, opened 2026-05-22
Lazy refresh causes retry storm of CLUSTER SHARDS/SLOTS


Luis729:
When a client needs to refresh its cluster topology, it does a lazy refresh. This comes from tcp timeout, a clusterdown, etc...

https://github.com/valkey-io/valkey-go/blob/873127484c04836551183b675ac6e4ac967f97d7/cluster.go#L155-L157

We had an incident before where we had ~5000 valkey-go clients connect to a 1000-shard Valkey cluster (1000 primaries, no replicas). This is approximately 5 million connections total. When we had an automation reduce the 1000 shard to 500 shards, we ran CLUSTER RESET SOFT on the old 500 shards, which caused a CLUSTERDOWN event to the valkey-go client. The CLUSTERDOWN event caused a topology refresh from each valkey-go client.

With the size of CLUSTER SHARDS (or even CLUSTER SLOTS for Redis 7 and below), and # of commands being called in a second, our cluster CPU usage spiked to 100%.

<img width="751" height="154" alt="Image" src="https://github.com/user-attachments/assets/c3e4cc32-ef7b-44ac-9584-dbbef1ed4028" />

(there are two graphs in the cluster shards, one is delayed for ~5m while the other is ~1m but same metric)
<img width="747" height="224" alt="Image" src="https://github.com/user-attachments/assets/988fe96d-02b1-4921-a8d2-5977f8f76077" />

I created an internal fork that configures how fast we refresh as a mitigation but I think its important to have this as an issue to see if this is a bigger fix or not. I'll make a PR for that change in a bit.

But the problems we saw was:

Refresh every second
https://github.com/valkey-io/valkey-go/blob/873127484c04836551183b675ac6e4ac967f97d7/cluster.go#L156

(I think its minor) Batching every 4 connections
https://github.com/valkey-io/valkey-go/blob/873127484c04836551183b675ac6e4ac967f97d7/cluster.go#L199

Immediately probing next connection if results of the current one give []/error
https://github.com/valkey-io/valkey-go/blob/873127484c04836551183b675ac6e4ac967f97d7/cluster.go#L208-L210

rueian:
Hi @Luis729, thanks for raising the issue. Regarding your graph, what are these 2 lines?

<img width="371" height="86" alt="Image" src="https://github.com/user-attachments/assets/8478e1ae-1387-4585-886e-ac9432b2b9ed" />

Luis729:
[...]

PR #141, opened 2026-05-22
Distribute # of calls to server during topology refresh

Luis729:
See https://github.com/valkey-io/valkey-go/issues/140 for info

This PR introduces default topology refresh behavior, reducing the storm of CLUSTER SHARD/SLOT calls to the Valkey server when the refresh happens.

## On close

Before: The client sends CLUSTER SHARDS call to all nodes 1 second after connection is closed.

After: The client sends CLUSTER SHARDS call to nodes [0,n] seconds after connection is closed. n depends on # of nodes the client is connected to, starting at 1 second for 100 nodes, and up to 30 seconds for 1000+ nodes.

## Batch calls

Before: The client sends CLUSTER SHARDS calls to all nodes at the same time (after the 1 second has passed). There is no delay between calls.

After: The client has delay between calls from 0 to 100ms. It is zero for less than 100 nodes, and up to 100ms for 1000+ nodes.

## Prefer init address on refresh

To further reduce the number of calls made, we introduce `ClusterOptions.PreferInitAddressRefresh` (false by default) which can be set if we can trust the addresses provided on `InitAddress` to exist after a topology refresh. It will call connections to all nodes in `InitAddress` on refresh.

## Testing

Ran two clusters at smaller scale (400 primaries, zero replicas), and had two 1000 k8 pods - one with valkey-go and other with rueidis client - doing 50 HMGET/s to them. Then I ran `systemctl stop valkey-server` for 5 minutes, and started it after.

valkey-go w/ this change (EnableClusterRefreshSpread enabled)
image

rueidis:
image

----

I do ask for feedback on this, I think this is necessary but wanted to see community's opinion on the numbers we use.

Because this information is very specific valkey-go information, I wanted to internalize a lot of this by using # of nodes the client connected to, but I am not strongly opinionated on what should stay in and what should be exposed.
   - delay between each refresh (1 to 30 seconds if EnableClusterRefreshSpread enabled)
   - delay between batch search (0 to 100ms if EnableClusterRefreshSpread enabled)
   - number of connections to batch per iteration (left it at 4)

I do know that with 1 second refresh time, and no rate limit on checking topology, we will see this storm happen with really large clusters.

-----

Old description:

~~This PR introduces two knobs, whose defaults are what valkey-go uses today (so no changes to existing code, fully backwards compatible).~~

- ~~EnableClusterRefreshSpread: (default: false) when enabled, the topology refresh is distributed in a random period of time, which we configure based on the size of the cluster.~~
    - ~~n< 100 nodes: 1 second~~
    - ~~100≤n<1000: linear from 1 to 30 seconds~~
    - ~~n≥1000: 30 seconds~~

- ~~ClusterRefreshMaxDelay: a separate max distribution (will override the 1-30 seconds if set)~~
----

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -229,37 +229,15 @@ func getClusterSlots(c conn, timeout time.Duration) clusterslots {
 }
 
 func (c *clusterClient) _refresh() (err error) {
-	c.mu.RLock()
-	results := make(chan clusterslots, len(c.conns))
-	pending := make([]conn, 0, len(c.conns))
-	for _, cc := range c.conns {
-		pending = append(pending, cc.conn)
-	}
-	c.mu.RUnlock()
-
-	var result clusterslots
 	batchDelay := c.clusterRefreshBatchDelay()
-	for i := 0; i < cap(results); i++ {
-		if i&3 == 0 { // batch CLUSTER SLOTS/CLUSTER SHARDS for every 4 connections
-			if i > 0 && batchDelay > 0 {
-				time.Sleep(batchDelay)
-			}
-			for j := i; j < i+4 && j < len(pending); j++ {
-				go func(c conn, timeout time.Duration) {
-					results <- getClusterSlots(c, timeout)
-				}(pending[j], c.opt.ConnWriteTimeout)
-			}
-		}
-		result = <-results
-		err = result.reply.Error()
-		if len(result.reply.val.values()) != 0 {
-			break
-		}
+	preferred, fallback := c.clusterRefreshConns()
+	result, err := c.refreshConns(preferred, batchDelay)
+	if len(result.reply.val.values()) == 0 && len(fallback) > 0 {
+		result, err = c.refreshConns(fallback, batchDelay)
 	}
 	if err != nil {
 		return err
 	}
-	pending = nil
 
 	groups := result.parse(c.opt.TLSConfig != nil)
 	conns := make(map[string]connrole, len(groups))
@@ -384,6 +362,55 @@ func (c *clusterClient) _refresh() (err error) {
 	}
 
 	return nil
+}
+
+func (c *clusterClient) clusterRefreshConns() (preferred, fallback []conn) {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	if !c.opt.ClusterOption.PreferInitAddressRefresh {
+		fallback = make([]conn, 0, len(c.conns))
+		for _, cc := range c.conns {
+			fallback = append(fallback, cc.conn)
+		}
+		return nil, fallback
+	}
+
+	initAddrs := make(map[string]struct{}, len(c.opt.InitAddress))
+	for _, addr := range c.opt.InitAddress {
+		initAddrs[addr] = struct{}{}
+		if cc, ok := c.conns[addr]; ok {
+			preferred = append(preferred, cc.conn)
+		}
+	}
+	fallback = make([]conn, 0, len(c.conns))
+	for addr, cc := range c.conns {
+		if _, ok := initAddrs[addr]; !ok {
+			fallback = append(fallback, cc.conn)
+		}
+	}
+	return preferred, fallback
+}
+
+func (c *clusterClient) refreshConns(pending []conn, batchDelay time.Duration) (result clusterslots, err error) {
+	results := make(chan clusterslots, len(pending))
+	for i := 0; i < len(pending); i++ {
+		if i&3 == 0 { // batch CLUSTER SLOTS/CLUSTER SHARDS for every 4 connections
+			if i > 0 && batchDelay > 0 {
+				time.Sleep(batchDelay)
+			}
+			for j := i; j < i+4 && j < len(pending); j++ {
+				go func(c conn, timeout time.Duration) {
+					results <- getClusterSlots(c, timeout)
+				}(pending[j], c.opt.ConnWriteTimeout)
+			}
+		}
+		result = <-results
+		err = result.reply.Error()
+		if len(result.reply.val.values()) != 0 {
+			break
+		}
+	}
+	return result, err
 }
 
 func (c *clusterClient) single() (conn conn) {
--- a/cluster.go
+++ b/cluster.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"io"
+	"math/rand"
 	"net"
 	"strconv"
 	"sync"
@@ -100,9 +101,6 @@ func newClusterClient(opt *ClientOption, connFn connFn, retryer retryHandler) (*
 		return nil, err
 	}
 
-	if d := client.clusterRefreshStartDelay(); d > 0 {
-		time.Sleep(d)
-	}
 	if err := client.refresh(context.Background()); err != nil {
 		return client, err
 	}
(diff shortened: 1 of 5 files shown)


Luis729:
My opinion btw:
 
We could completely internalize this fwiw but that's going to be a behavior change. Using # nodes is good to determine threshold. What I think is required vs what I think is minor:

Most important:
- spreading out _refresh calls over period of time, because if 5000 clients exist to 1000 clients, that's a lot of calls per client after that 1 second mark

Still important:
- delay before next iteration: so if CLUSTER SHARDS/SLOTS give [], we should wait a small duration before calling next set of connections. If 500/1000 nodes are down, at worst case scenario, the client can send out 501 CLUSTER SHARDS request in one small iteration. With 5000 clients, thats a LOT of calls.

Of course, the risk here is that some GET might lag while topology is refreshing, so I understand why 1 second and why immediate.

rueian:
Hi @Luis729,

I think we definitely need some changes to how we refresh the topology. But the 2 knobs, in my opinion, are too hard to configure. I would expect something that is more out of the box and is more beneficial to all cluster sizes.

May we just try:
1. Instead of batching CLUSTER SLOTS/CLUSTER SHARDS for every 4 connections, batch for every 2 or 1 connection.
2. Ensure the randomization of the batch.
3. Exponential backoff + jitter on each batch.

Luis729:
Yeah I wasn't a fan of the knobs, but its what we're doing it within my company for now. I think something out of the box is better long term so I can look to doing this when I get the time, should be soon.

But there is a risk. When the topology is not refreshed, of course a GET to that node will return an error (tcp err, clusterdown, etc) and so I'm not sure how the community here is planning on pushing that out (will it be a minor version vs patch change? cause technically this is a behavioral change and everyone on valkey-go should expect additional time to refresh topology)

For example, what I wrote in one of our docs internally (redacted ofc)

-----

We suggest the following configs to reduce the spike in CLUSTER SHARDS:
- ClusterRefreshThreshold to 30 seconds
- ClusterRefreshJitter to 100% (so it refreshes [0,30] seconds)
- ClusterRefreshConnConcurrency to 1 (4 is OK though I think, but 1 is safer).
- ClusterRefreshIntraDelay to 100ms

We estimate with a drop from 1000 -> 500 nodes and with these configs:
1. First topology check [0, 30] random
2. If the check is empty, wait 100ms so 500x that is 50 seconds

We can see a delay of approximately 1-1.5min (versus an ideal 1-2 seconds if all the calls go through and no issues) to refresh topology. Until then, there will be some errors from the client stating the node is empty/no slots.

(the ideal 1-2 seconds didnt happen btw, what happened was CPU spiked up and we had to deal with this for 2.5h hours ahahaa)

rueian:
Another idea is allowing you to specify a list of nodes that you want valkey-do to refresh the topology with. Otherwise, in an event like this, scaling down half the cluster, you will have unavoidably 50% of chance to fail to refresh the topology.

Luis729:
> Another idea is allowing you to specify a list of nodes that you want valkey-go to refresh the topology with. Otherwise, in an event like this, scaling down half the cluster, you will have unavoidably 50% of chance to fail to refresh the topology.

I will take a look into this too, I was really busy in the past week but I should have time to get something out this week. But this idea can be an option of course. I cant predict everyone will use the same node for their cluster (like for example, I can probably convince my team to use a node in shard 0, but this has plenty of edge cases to consider)

rueian:
>  I cant predict everyone will use the same node for their cluster

Nodes listed in the `InitAddress` should be more stable than the others because those nodes are the only nodes the client can connect to when it is restarted.

So, probably we just need to add a bool option in the `ClusterOption` to let the client only fetch the topology from those nodes.

Luis729:
@rueian thanks for the feedback, sorry for the overdue change. Let me know what you think. I also updated PR description. We can stick to the 30 second max for starting topology refresh and 100ms second max for batch.

Luis729:
btw I left the # batch at 4. I dont think it is significant to change it to 2 or 1. I think what is signficant is just sending all of them at once.

cluster.go
```diff
@@ -100,6 +100,9 @@ func newClusterClient(opt *ClientOption, connFn connFn, retryer retryHandler) (*
 		return nil, err
 	}
 
+	if d := client.clusterRefreshStartDelay(); d > 0 {
+		time.Sleep(d)
+	}
```
rueian: We probably don't want to sleep at the initialization.

cluster.go
```diff
@@ -339,6 +364,55 @@ func (c *clusterClient) _refresh() (err error) {
...
+			preferred = append(preferred, cc.conn)
+		}
+	}
+	fallback = make([]conn, 0, len(c.conns))
+	for addr, cc := range c.conns {
+		if _, ok := initAddrs[addr]; !ok {
+			fallback = append(fallback, cc.conn)
+		}
+	}
+	return preferred, fallback
```
rueian: I don't think the fallback is needed. Let's keep it simple: if PreferInitAddressRefresh is set, then only use InitAddress.

cluster.go
```diff
@@ -339,6 +364,55 @@ func (c *clusterClient) _refresh() (err error) {
...
 
+func (c *clusterClient) clusterRefreshConns() (preferred, fallback []conn) {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	if !c.opt.ClusterOption.PreferInitAddressRefresh {
+		fallback = make([]conn, 0, len(c.conns))
+		for _, cc := range c.conns {
+			fallback = append(fallback, cc.conn)
+		}
+		return nil, fallback
```
rueian: Do a rand.Shuffle to ensure randomness.

cluster.go
Luis729: Updated PR description to reflect

cluster.go
Luis729: Updated - did shuffle on the initAddress, the map c.conns is already randomized by nature so no need to shuffle that again

rueian: > the map c.conns is already randomized by nature so no need to shuffle that again

I thought that too, but it turns out it is usually not random enough.

Luis729: I can add a shuffle here too, I know our existing behavior (what's in main rn) doesnt shuffle but just pointing that out

valkey.go
```diff
@@ -310,6 +310,9 @@ type ClusterOption struct {
 	// Cluster topology cache refresh happens always in the background after a successful scan.
 	ShardsRefreshInterval time.Duration
 
+	// PreferInitAddressRefresh only uses ClientOption.InitAddress nodes during cluster topology refresh.
+	PreferInitAddressRefresh bool
```
rueian: This should be moved to after `MaxMovedRedirections`.

cluster.go
```diff
@@ -153,7 +154,47 @@ func (c *clusterClient) refresh(ctx context.Context) (err error) {
 }
 
 func (c *clusterClient) lazyRefresh() {
-	c.sc.LazyDo(time.Second, c._refresh)
+	c.sc.DelayDo(c.clusterRefreshStartDelay(), c._refresh)
+}
+
+func (c *clusterClient) clusterRefreshStartDelay() time.Duration {
+	maxDelay := c.clusterRefreshMaxDelay()
+	return time.Duration(util.FastRand(int(maxDelay))) + time.Nanosecond
```
```suggestion
	return time.Duration(util.FastRand(int(maxDelay)))
```

Adding one nanosecond doesn't seem to be helpful.

Luis729: 🤦

rueian (approved):
Thanks @Luis729

PR #142, opened 2026-05-25
feat: standalone failover detection and periodic replica refresh

mhr3:
## Summary

Makes the standalone client topology-aware so it can survive primary failovers without intervention, and adds a new `Standalone.ReplicaRefreshInterval` option to periodically reconcile the replica pool. The motivating use case is AWS ElastiCache / MemoryDB in non-cluster mode where:

- the primary endpoint DNS flips on failover,
- a reader endpoint cannot be relied on for AZ-aware reads (it round-robins across replicas on every resolve), so callers want to list each replica's *per-node* DNS in `ReplicaAddress`,
- any of those per-node DNSes may end up promoted to primary at any time.

Today, listing a per-node DNS that gets promoted breaks the existing standalone client — there is no failover handling and the connection is locked to whichever role it started as. This PR fixes that without requiring `EnableRedirect` (which is Valkey 8+ only and not supported by all managed providers).

### Behaviour

- **Role discovery is free.** `conn.Role()` exposes the role label already cached from HELLO, so init can identify the primary among the configured `ReplicaAddress` entries with no extra round trip and silently dedupe any entry that turns out to be the primary.
- **Reactive failover.** When a write hits a node that has been demoted (`READONLY` error), the client runs `ROLE` across the primary and every known replica, promotes whichever one now reports as master, swaps the routing atomically, and retries the original command. This path is always on; no new option needs to be set.
- **Optional periodic reconcile.** When `Standalone.ReplicaRefreshInterval > 0` a background goroutine runs the same reconcile at the configured interval. In addition to promoting on failover, it:
  - drops replicas whose `ROLE` shows state ≠ `connected` from the read pool,
  - re-dials addresses whose connection was previously dropped,
  - flips init into a lenient mode where dial failures and entries that resolve to the primary are silently skipped instead of aborting startup.
- **New helper.** `ValkeyError.IsReadOnly()` mirrors the existing `IsRedirect()` / `IsMoved()` / `IsAsk()` predicates and is used by the failover path; it's also useful to callers writing their own retry logic.

### API

One new field on `StandaloneOption` and one new method on `ValkeyError`:

```go
type StandaloneOption struct {
    ReplicaAddress         []string
    ReplicaRefreshInterval time.Duration // new
    EnableRedirect         bool
}

func (r *ValkeyError) IsReadOnly() bool // new
```

`ReadNodeSelector` / `EnableReplicaAZInfo` already exist and now work as documented for standalone deployments too — see the new README example.

### Example

```go
client, err := valkey.NewClient(valkey.ClientOption{
    InitAddress: []string{"primary-endpoint.cluster.cache.aws:6379"},
    Standalone: valkey.StandaloneOption{
        ReplicaAddress: []string{
            "node-001.cluster.cache.aws:6379",
            "node-002.cluster.cache.aws:6379",
            "node-003.cluster.cache.aws:6379",
        },
        ReplicaRefreshInterval: 5 * time.Second,
    },
    SendToReplicas: func(cmd valkey.Completed) bool {
        return cmd.IsReadOnly()
    },
    EnableReplicaAZInfo: true,
    ReadNodeSelector:    valkey.AZAffinityNodeSelector("us-east-1a"),
})
```

### Compatibility

- No breaking changes to existing call sites. `Standalone.ReplicaAddress` with default `ReplicaRefreshInterval` (zero) keeps the previous strict-init semantics: dial errors and per-node DNS that turns out to be the primary still cause init to fail. The only behavioural difference is that reactive READONLY-driven failover is now always on; this is purely additive — a write that previously bubbled READONLY up to the caller will now retry against the new primary instead.
- `Standalone.EnableRedirect` is unchanged.

## Tests

- New unit tests in `standalone_test.go` covering: dedup-at-init via `Role()`, READONLY-triggered reconcile and primary promotion, dropping disconnected slaves, re-dialing previously-dropped addresses, and the periodic monitor loop (including a "promote new primary purely from the monitor" case with no traffic).
- New tests for `ValkeyError.IsReadOnly()` in `message_test.go`.
- All existing standalone tests updated to the new mock surface (`mockConn.RoleFn`).

## Test plan

- [x] `go test -short ./...`
- [x] Built a small failover harness against ElastiCache / MemoryDB and validated:
  - steady-state reads distribute across the configured per-node DNS replicas,
  - manually triggered primary failover causes a few seconds of writer errors followed by a clean "recovered" log line, with reads continuing throughout,
  - the periodic monitor reflects the new topology in `Nodes()` even when no traffic hits the demoted node.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/standalone.go
+++ b/standalone.go
@@ -2,13 +2,42 @@ package valkey
 
 import (
 	"context"
+	"errors"
 	"maps"
 	"math/rand/v2"
+	"slices"
+	"sync"
 	"sync/atomic"
 	"time"
 
 	"github.com/valkey-io/valkey-go/internal/cmds"
 )
+
+const masterRole = "master"
+
+var errNoPrimaryFound = errors.New("no primary found")
+
+type standalone struct {
+	retryer        retryHandler
+	toReplicas     func(Completed) bool
+	nodeSelector   func(uint16, []NodeInfo) int
+	primary        atomic.Pointer[singleClient]
+	state          atomic.Pointer[standaloneState]
+	connFn         connFn
+	opt            *ClientOption
+	redirectCall   call
+	reconcileCall  call
+	stop           chan struct{}
+	stopOnce       sync.Once
+	enableRedirect bool
+}
+
+type standaloneState struct {
+	masterAddr string // prioritized master address from ReplicaAddress
+	clients    map[string]*singleClient
+	nodes      []NodeInfo // length == len(replicas)+1; nodes[0] mirrors the current primary
+	replicas   []*singleClient
+}
 
 func newStandaloneClient(opt *ClientOption, connFn connFn, retryer retryHandler) (*standalone, error) {
 	if len(opt.InitAddress) == 0 {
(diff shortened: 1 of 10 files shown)


standalone.go
```diff
@@ -107,61 +195,205 @@ func (s *standalone) redirectToPrimary(addr string) error {
...
+			return recErr == nil || s.retryer.WaitOrSkipRetry(ctx, attempts, cmd, err)
+		}
+	}
+	if s.opt.Standalone.ReplicaRefreshInterval > 0 && valkeyErr.IsReadOnly() {
+		prev := s.primary.Load()
+		_ = s.reconcile(ctx)
+		if s.primary.Load() != prev {
+			return true
 		}
+		return s.retryer.WaitOrSkipRetry(ctx, attempts, cmd, err)
```
rueian, line 224: I don't think it is safe to retry write commands silently. We'd better still return the error to the user.

rueian:
>  silently dedupe any entry that turns out to be the primary.

I am not sure if this is a good idea. It seems like doing too much. What if somehow some users want to keep them duplicated?

The current `reconcile` function looks a bit complicated. Can't we just send a `ROLE` command to the old primary node to find the location of the new master when we see a `READONLY` error? We probably don't need to send ROLE to every node periodically.

BTW, the DCO check doesn't pass.

mhr3:
> > silently dedupe any entry that turns out to be the primary.
> 
> I am not sure if this is a good idea. It seems like doing too much. What if somehow some users want to keep them duplicated?
> 

The reason for doing this is that in this mode the []ReplicasAddress is expected to list every node, but at the same time assumes the primary (opt.InitAddress[0]) is different than any of the node addresses - despite "pointing" to one of the nodes.

> The current `reconcile` function looks a bit complicated. Can't we just send a `ROLE` command to the old primary node to find the location of the new master when we see a `READONLY` error? We probably don't need to send ROLE to every node periodically.

I'm actually thinking about removing the READONLY part of this. We've been running this for a while, and have never seen it trigger - on failover the connections usually fail with some kind of net.Error, and after the DNS updates, new connections work fine. So actually the periodic ROLE commands is what makes this work - both for removing replicas that are no longer synced and to update things when a replica becomes primary.

rueian:
> on failover the connections usually fail with some kind of net.Error

What kind of net errors did you see? Did you also see net errors on replicas?

mhr3:
> > on failover the connections usually fail with some kind of net.Error
> 
> What kind of net errors did you see? Did you also see net errors on replicas?

IIRC there were mostly EOF errors, dial timeouts and i/o timeouts in general. Did not see a replica failure so far, but I imagine it would be similar.

rueian:
> The reason for doing this is that in this mode the []ReplicasAddress is expected to list every node, but at the same time assumes the primary (opt.InitAddress[0]) is different than any of the node addresses - despite "pointing" to one of the nodes.

Yes, but I still hesitate to do the duplication because `ReplicaAddress` is explicitly set by users, and I think silently dropping one of them may surprise users. For example, a user may just assume NodeInfo[0] is always the primary and NodeInfo[1] is always the node-001, NodeInfo[2] is always the node-002, etc.

Maybe providing the `run_id` from the `INFO` response in the `NodeInfo` struct is enough.

mhr3:
My initial version was adding a new `[]string NodeAddress` field into the Standalone struct instead of using the `ReplicaAddress`, but given you wouldn't really be able to use both at the same time, I decided against it. But perhaps that would make things better - afterall you're not really passing in replica addresses.

mhr3:
> For example, a user may just assume NodeInfo[0] is always the primary and NodeInfo[1] is always the node-001, NodeInfo[2] is always the node-002, etc.

Also that goes against the goal here, if a replica is down / stopped syncing with master, we remove it from the list (temporarily). Perhaps I need to make the docs be more explicit about that.

PR #143, opened 2026-06-09
opt-in static-TTL client-side cache (skip MULTI/PTTL/EXEC, 2-cmd wire)

sriramtallapragada:
### Summary

Adds `Cacheable.StaticTTL()` — a per-command builder flag that makes the caller-provided `ttl` argument authoritative for the L1 entry's lifetime in `DoCache` / `DoMultiCache`. When set, the wire is reduced from the standard 5-cmd transactional shape `[CLIENT CACHING YES, MULTI, PTTL, cmd, EXEC]` to a 2-cmd wire `[CLIENT CACHING YES, cmd]`. The cacheable reply is committed to L1 directly by the connection's read goroutine via a new internal tag bit (`staticTTLTag`).

For `DoMultiCache`, the 2-cmd direct wire is taken only when **every** command in the batch is tagged with `StaticTTL`; a batch containing any untagged command falls back to the standard MULTI/PTTL/EXEC wire for all of them.

Default path (untagged commands) is byte-for-byte unchanged.

### Why explore this

We observed that on production workloads the MULTI/PTTL/EXEC wrapper around CSC reads consumes a large fraction of server-side CPU. Pulling per-command CPU time from one primary node over 153.9 days uptime
(via cumulative `INFO commandstats`):

| # | Command | Total CPU (µsec) | Calls | µsec/call | % of cmd time |
|---|---|---:|---:|---:|---:|
| 1 | **exec** | 321,615,160,895 | 76,727,665,777 | 4.19 | **61.96%** |
| 2 | **pttl** | 55,000,607,242 | 80,226,985,663 | 0.69 | **10.60%** |
| 3 | evalsha | 44,313,556,501 | 3,789,203,430 | 11.69 | 8.54% |
| 4 | setex | 19,941,353,011 | 5,474,103,481 | 3.64 | 3.84% |
| 5 | get | 15,336,559,245 | 70,760,054,410 | 0.22 | 2.96% |
| 6 | **client\|caching** | 13,202,217,403 | 76,727,662,050 | 0.17 | **2.54%** |
| 7 | hmget | 6,633,126,391 | 8,508,069,266 | 0.78 | 1.28% |

The three commands the patch eliminates per cacheable miss — `EXEC`, `PTTL`, `CLIENT CACHING YES` — together account for **~75% of cumulative command-execution time** on this node. `EXEC` alone is
~62%. The call counts for `EXEC` and `CLIENT CACHING YES` match exactly: every CSC read paid for one EXEC.

The PTTL clamp exists to bound L1 staleness against Valkey's lazy-expiry semantics — a key past its TTL stays readable in memory until something triggers deletion, and only deletion fires a CLIENT TRACKING push. For callers whose freshness is enforced by an active invalidation mechanism (writes that explicitly invalidate the affected
cached keys), the push fires on the explicit mutation and bounds staleness independent of the PTTL clamp. On those workloads, the wrapper is paying CPU for a guarantee the workload doesn't need.

### Production validation

We forked `valkey-go` to test this at scale. After deploying behind the new opt-in:

- **~72%** of cacheable reads now route through the optimized 2-cmd
  wire.
- **~55% reduction** in read-path commands at peak.
- **~9.85M commands/sec eliminated** at peak.

These are observed deltas from the production rollout.

#### Loopback Valkey (per-miss client-side cost)

Reproducible by any reviewer via the new `BenchmarkClientSideCachingMiss` benchmark, which forces every iteration to be an L1 miss by using a unique key per call (`atomic.Uint64` counter):

```
docker compose up -d valkey
go test -bench='ClientSideCachingMiss' -benchtime=10s -benchmem \
        -run=^$ -cpu=4 .
```

Apple M4 Max, loopback Valkey 8.0:

| Subtest | ops | ns/op | B/op | allocs/op |
|---|---:|---:|---:|---:|
| `DoCache` (standard) | 2.30M | 6576 | 974 | 18 |
| `DoCacheStaticClientTTL` | **3.28M** | **5071** | **794** | **14** |
| Δ | **+42% throughput** | **−23%** | **−18%** | **−22%** |

Hot-path L1 hits are unchanged (parity within noise on `BenchmarkClientSideCaching/DoCache` vs `/DoCacheStaticClientTTL`, both at ~66 ns/op, 0 allocs).

### What the patch changes

| File | Change |
|---|---|
| `internal/cmds/cmds.go` | New `staticTTLTag` flag bit; `Cacheable.StaticTTL()` builder; package-level `IsStaticTTL(Completed)` predicate; `ClearStaticTTL(Completed)` helper used by the ASK-redirect path. |
| `pipe.go` | New 2-cmd wire branch in `DoCache` (gated on the cmd's `StaticTTL` tag); matching branch in `DoMultiCache` (gated on **every** `CacheableTTL` in the batch carrying the tag); new `else if cmds.IsStaticTTL(multi[ff])` auto-commit gate in `_backgroundRead`. The gate is checked BEFORE the standard CSC gate — otherwise an array reply of length ≥ 2 on the cacheable cmd would incidentally match the standard gate. |
| `cluster.go` | `askingMultiCache` emits the matching 2-cmd ASK wire shape when every batched cmd is tagged. The tag is **stripped** from the wire via `cmds.ClearStaticTTL` — the redirected connection has no Flight slot to resolve, so the read loop's direct-commit gate must not fire. |

The reply commits to L1 on the connection's read goroutine, at the same point in the loop as the standard CSC auto-commit, so the invalidation race window is identical to today's.

### Behavioral divergence on the opt-in (intentional, pinned by tests)

| Wire condition | Standard CSC | `Cacheable.StaticTTL()` |
|---|---|---|
| Inner cmd success | val=reply, err=nil | val=reply, err=nil |
| Inner cmd nil reply | val=typeNull, err=nil (cached) | val=typeNull, err=nil (cached) |
| Inner cmd typed error (e.g. `-WRONGTYPE`) | cached with `cacheMark`; `IsCacheHit()`→true on next read | L1 slot cancelled; `IsCacheHit()`→false on next read, fresh wire trip |
| Transport error | newErrResult(err) | newErrResult(err) |

`result.Error()`, `result.NonValkeyError()`, and `result.ToMessage()` are identical on both paths. The one observable delta is
`IsCacheHit()` on follow-up reads of an erroring key. Arguably the safer behaviour — caching `-WRONGTYPE` for the full client TTL hides programming bugs under standard CSC. Pinned by `TestClientSideCachingStaticClientTTLCleanupNoRaceWithReadLoop/OnWireError`.

### Tests

- 8 new top-level tests + 5 subtests in `pipe_test.go`, following the
  existing `TestClientSideCaching*` convention.
- 1 new tag-mechanic test in `internal/cmds/cmds_test.go`
  (`TestIsStaticTTL`, covering `Cacheable.StaticTTL` setter and
  `ClearStaticTTL` helper).
- Plus 2 regression guards on `CacheKey`'s bitmask matching (already
  in `internal/cmds/cmds_test.go`).
- Race-clean `-count=3` on the full CSC test surface.
- Existing test suite unmodified and green.

### Out of scope for this PR

- **MGET / JSON.MGET fast path.** `DoCache` dispatches to `doCacheMGet` before consulting the `StaticTTL` tag, so MGET stays on the standard 5-cmd-per-key wire even when the cacheable is tagged. Documented in the `Cacheable.StaticTTL` Godoc. `DoMultiCache` rejects MGET outright. A parallel `doCacheMGetStaticClientTTL` could be added in a follow-up if there's demand.


rueian:
Cool. Thanks for the PR.

internal/cmds/cmds.go
```diff
@@ -134,6 +138,13 @@ func (c Completed) ToRetryable() Completed {
 	return c
 }
 
+// ToDirectCacheCommit returns a copy of this command with
+// directCacheCommitTag set. See the tag's definition.
+func (c Completed) ToDirectCacheCommit() Completed {
```
rueian:
```suggestion
func (c Cacheable) StaticTTL() Cacheable {
```

We don't need a new `WithStaticClientTTL` context helper. The tag is already carried in the command.

internal/cmds/cmds.go
```diff
@@ -144,6 +155,11 @@ func (c *Completed) IsOptIn() bool {
 	return c.cf&optInTag == optInTag
 }
 
+// IsDirectCacheCommit reports whether directCacheCommitTag is set.
+func (c *Completed) IsDirectCacheCommit() bool {
```
rueian:
```suggestion
func IsStaticTTL(cmd Completed) bool {
```

Keep the checker function private to users.

cache_static_client_ttl.go
```diff
@@ -0,0 +1,32 @@
...
+// MGET / JSON.MGET fall through to the standard wire shape; this opt-in
+// has no effect on them.
+func WithStaticClientTTL(ctx context.Context) context.Context {
+	return context.WithValue(ctx, staticClientTTLCtxKey{}, true)
+}
+
+func hasStaticClientTTL(ctx context.Context) bool {
+	v, _ := ctx.Value(staticClientTTLCtxKey{}).(bool)
+	return v
+}
```
rueian:
```suggestion
```
We don't need these.

For DoMultiCache, we can either require all commands to be with static TTL or at lease one of the commands to be with static TTL to enable the direct path.

internal/cmds/cmds.go
sriramtallapragada: Good point, removed the file.

internal/cmds/cmds.go
sriramtallapragada: Makes sense, [updated](https://github.com/valkey-io/valkey-go/pull/143/changes/29557b4d6aa4b94527cc1d44dff8353cedbe73f7#diff-658409170ce47d9bc259eeca31798fa4438cae04c6c2440fd0f3d0d855303351R155)

cache_static_client_ttl.go
sriramtallapragada: Removed them and as the starting point for DoMultiCache, I have added in to require all commands to be with static TTL (documented the same as well), this can be further improved as needed.

sriramtallapragada:
> Cool. Thanks for the PR.

Thanks for taking time to review, I have addressed your comments.

sriramtallapragada:
@rueian Added more tests to increase the coverage. Please review when you get a chance and let me know if any more changes are needed.

cluster.go
```diff
@@ -1054,14 +1088,25 @@ func (c *clusterClient) askingMultiCache(cc conn, ctx context.Context, multi []C
...
-			if preErr := resps.s[i-1].Error(); preErr != nil { // if {Cmd} get a ValkeyError
-				err = preErr
+	if skipMultiExec {
+		// Cmd reply is the user-visible result directly (no EXEC unwrap).
+		for i := offset; i < len(resps.s); i += stride {
+			if resps.s[i].err != nil {
+				results.s = append(results.s, newErrResult(resps.s[i].err))
+			} else {
+				results.s = append(results.s, newResult(resps.s[i].val, nil))
+			}
```
```suggestion
			results.s = append(results.s, resps.s[i])
```

pipe.go
```diff
@@ -621,6 +621,22 @@ func (p *pipe) _backgroundRead() (err error) {
 			if multi == nil {
 				multi = ones
 			}
+		} else if cmds.IsStaticTTL(multi[ff]) {
```
rueian: Could you help check if this `cmds.IsStaticTTL(multi[ff])` is inlined? Otherwise, the function call that copies the command struct could be too expensive.

pipe.go
```diff
@@ -1629,12 +1657,28 @@ func (p *pipe) DoMultiCache(ctx context.Context, multi ...CacheableTTL) *valkeyr
 			panic(panicmgetcsc)
 		}
 	}
+	// All-or-nothing: only take the stride-2 direct path when every
+	// CacheableTTL in the batch was tagged with .StaticTTL(). A mixed
+	// batch falls back to the standard stride-5 wrapped wire.
+	skipMultiExec := true
+	for _, ct := range multi {
+		if !cmds.IsStaticTTL(Completed(ct.Cmd)) {
```
rueian, line 1666: Same here. Is this inlined?

pipe.go
sriramtallapragada: Verified — inlined at every call site, no struct copy on the hot path.

```
./pipe.go:624:29:    inlining call to cmds.IsStaticTTL
  ./pipe.go:1496:21:   inlining call to cmds.IsStaticTTL
  ./pipe.go:1665:23:   inlining call to cmds.IsStaticTTL
  ./cluster.go:1043:23: inlining call to cmds.IsStaticTTL
  ./cluster.go:1061:82: inlining call to cmds.ClearStaticTTL

  Reproducible with:
  go build -gcflags='-m' ./... 2>&1 | grep -E 'inlining call to cmds\.(IsStaticTTL|ClearStaticTTL)'
```

pipe.go
sriramtallapragada: yes it's inline here as well.

cluster.go
sriramtallapragada: Good point, updated it [here](https://github.com/valkey-io/valkey-go/pull/143/changes/0c18b3c5a7827a3f0e14aaceb86e21e332a884ca#diff-556b0079cdcdaaa712eaa26594117c9d72b568bbda914495d8f46ee847f98c40R1094)

cluster.go
```diff
@@ -1031,20 +1031,54 @@ func (c *clusterClient) askingMulti(cc conn, ctx context.Context, multi []Comple
...
+	)
+	if skipMultiExec {
+		// Wire per key: [OPT_IN, ASKING, cmd]. The staticTTLTag is
+		// stripped — this connection has no Flight slot for the key,
+		// so the read loop's direct-commit gate must not fire.
+		stride = 3
+		offset = 2
+		commands = make([]Completed, 0, len(multi)*stride)
+		for _, cmd := range multi {
+			commands = append(commands, cc.OptInCmd(), cmds.AskingCmd, cmds.ClearStaticTTL(Completed(cmd.Cmd)))
```
rueian: Curious why ClearStaticTTL is needed here? Do we need to do something similar to the !skipMultiExec path and the RedirectMove path?

pipe.go
rueian:
```suggestion
		} else if ff == 1 && cmds.IsStaticTTL(multi[ff]) {
```
To save some IsStaticTTL checks.

internal/cmds/cmds.go
```diff
@@ -213,6 +234,32 @@ func (c Cacheable) Pin() Cacheable {
...
+// MGET / JSON.MGET commands fall back to the default client-side
+// cache wire shape, which closely follows the Valkey-side key TTL
+// via the per-read PTTL clamp. DoCache dispatches them to a
+// dedicated multi-key path before this tag is consulted; DoMultiCache
+// rejects MGET outright.
+//
+// For DoMultiCache, the direct wire is taken only when every command
+// in the batch is tagged with StaticTTL. A batch containing any
+// untagged command falls back to the standard wire for all of them.
+func (c Cacheable) StaticTTL() Cacheable {
```
rueian:
```suggestion
func (c Cacheable) ToStaticTTL() Cacheable {
```
I reconsidered the method name. `ToStaticTTL` seems to be better.

cluster.go
sriramtallapragada: **Context on why ClearStaticTTL was here**: in the first iteration the opt-in lived on the context (`WithStaticClientTTL`), so all-or-nothing was implicit per call. After moving it to the command (`Cacheable.StaticTTL`), the all-or-nothing rule still decides the wire shape in DoMultiCache, but the per-cmd tag bits travel untouched. A mixed batch (some cmds tagged, some not) routes through the wrapped stride-5 wire with the tagged cmds still carrying staticTTLTag.

**That's a bug** — pinned now by `TestClientSideCachingStaticClientTTLDoMultiCacheMixedNoPollution`. Inside `MULTI/EXEC` the cmd reply at the cmd offset is "QUEUED". The static-TTL gate fires on it, calls `cache.Update(ck, cc, "QUEUED") `and resolves the Flight slot. The standard CSC gate at the EXEC offset later runs `cache.Update(ck, cc, real_value)` — but `lru.Update` only writes when the entry is still pending; a resolved entry is a silent no-op. So the cache holds "QUEUED" for the full TTL.

**Fix:** move ClearStaticTTL to the !skipMultiExec wrapped-wire build sites.
  - `pipe.DoMultiCache !skipMultiExec (both LRU and non-LRU loops)` — strip on every wire-side cmd.
  - `askingMultiCache !skipMultiExec` — same. re.cAskings collects originals with tags intact, so a mixed -ASK subset would hit the same bug here.
 - `askingMultiCache stride-3 branch (this comment)` — drop the strip. All cmds in this branch are tagged and letting the gate fire pre-warms the ASK-target's cache (matches standard CSC on -ASK).

`RedirectMove doesn't need anything analogous:` case RedirectMove runs a fresh ncc.DoCache(ctx, cmd, ttl) on the new owner, which builds its own wire on its own cache — exactly what we want, since ncc is the rightful new owner. (Added a test related to that as well)

**Also added the info on godocs:** 
// A batch containing any untagged command falls back to the standard wire for all of them
// (the all-or-nothing rule). For best results, prefer batches that
// are either fully tagged or fully untagged.

Fixed in this [commit](https://github.com/valkey-io/valkey-go/pull/143/changes/961deb30fdd8981e87c26201821cc068ad8676e8)

pipe.go
sriramtallapragada: Thanks, however I think the ff == 1 short-circuit might not always work, because cacheable cmds don't always land at ff=1? 

 `DoMultiCache` concatenates N keys into one `p.DoMulti(ctx, missing...)`, so for a stride-2 batch of N tagged keys the wire is [OPT_IN, cmd1, OPT_IN, cmd2, ..., OPT_IN, cmdN] — len(multi) = 2N and the cacheables sit at every odd ff:

```
    ff=0  OPT_IN ack          ff=2  OPT_IN ack              ff=4  OPT_IN ack    ...
    ff=1  cmd1 reply  ←gate   ff=3  cmd2 reply ←gate        ff=5  cmd3 reply ←gate
```

 With `ff == 1` only `cmd1` would hit the gate; `cmd2..cmdN's` Flight slots would never resolve and `entry.Wait(ctx)` would deadlock.  Verified by applying the suggestion and running the existing `TestClientSideCachingStaticClientTTLDoMultiCache/Hit` — it hangs to the 30s timeout.

 The "every odd ff" version (ff & 1 == 1) is correct for stride-2, but it then conflicts with the `stride-3 ASK wire` [OPT_IN, ASKING, cmd] × N, where cmds land at ff = 2, 5, 8, … — a mix of odd and even — so the gate would fire for some keys and skip others.

internal/cmds/cmds.go
sriramtallapragada: Good point, updated as part of this [commit](https://github.com/valkey-io/valkey-go/pull/143/changes/7c477670716e88f9659bd58b9a66348fd4a380e8).

pipe.go
rueian: Oh, you are right. Then at least do ff > 0?

```suggestion
		} else if ff > 0 && cmds.IsStaticTTL(multi[ff]) {
```

sriramtallapragada: Makes sense, [updated](https://github.com/valkey-io/valkey-go/pull/143/changes/f24d49edef6f732d6c19f694793cf48daf6a8a7d).

rueian (approved):
Thanks @sriramtallapragada!

Issue #146, opened 2026-07-09
Add Valkey Helpers for creating ValkeyResults


cfilby:
We're implementing an instrumented wrapper for Valkey-go and would like to include some opinionated defaults for users at our organization. One thing we've noticed is the ValkeyResponse type is public, but it's internal fields are private, which means that external modules have to use unsafe pointer conversions that could break if modules get out of sync. 

Is there a willingness to expose some constructors from valkey-go core like the following:

```
NewValkeyResponse(msg ValkeyMessage) ValkeyResponse
NewValkeyErrorResponse(err error) ValkeyResponse
```

This would allow us to remove unsafe pointer conversions from other modules like valkeyhooks/mock, and would minimize the risk of pointer conversions breaking if modules get out of sync. 

I'd be happy to contribute this, but wanted to check with y'all before opening a PR, I wasn't clear what the original motivation for the pointer conversions was.

rueian:
LGTM. Maybe the function names can be shorter, like NewResponse and NewErrorResponse.

cfilby:
Thank you! I'll try to get a PR open for this this week

cfilby:
@rueian I've taken a first pass at exposing and leveraging those constructors in #147. I didn't see any contribution guidelines, so my apologies if I'm missing anything.

rueian:
Merged. Thanks @cfilby.

PR #147, opened 2026-07-13
Expose and Use Constructors for ValkeyResults rather than Unsafe Pointer Conversions

cfilby:
## Description

Addresses #146 

Valkey-go doesn't directly provide a way for consumers of the library to instantiate the ValkeyErrorResult directly, which results in submodules performing unsafe pointer conversions. The main concern is if modules get out of sync and types like ValkeyResult or ValkeyResultStream evolve, this could lead to memory corruption/app crashes.

This PR:

* Exposes the NewResult constructor for instantiating ValkeyResults with a ValkeyMessage and error
* Exposes the NewErrorResult constructor for instantiating ValkeyResults with an underlying error
* Exposes a new NewErrorResultStream method for instantiating ValkeyResultStreams with an underlying error, and updates existing tests that return just an error to use it.
* Updates valkeyhook to remove duplicated types with unsafe pointer conversions. 
* Updates valkeycompat to use NewResult and NewErrorResult.
* Removes mock's result type since it can just use the newly exposed constructors.

The diff here is pretty large, but the bulk of these are refactorings or struct initializations being swapped for new constructors.

## Design Questions:

* NewResult(msg, err) vs NewResult(msg) - I opted for the former since a lot of the submodules seem to do direct result{msg, err} type calls. That said, this does create some duplicated API Surface Area with NewErrorResult.
* Mark Methods as Deprecated - The valkeyhook.NewErrorResponse could be marked as deprecated, as can some of the mock methods. This feels like a maintainer preference question, so opted to leave them as is for now.

### DIFF
--- a/client.go
+++ b/client.go
@@ -272,7 +272,7 @@ func (c *dedicatedSingleClient) Do(ctx context.Context, cmd Completed) (resp Val
 	attempts := 1
 retry:
 	if err := c.check(); err != nil {
-		return newErrResult(err)
+		return NewErrResult(err)
 	}
 	resp = c.wire.Do(ctx, cmd)
 	if c.retry && cmd.IsRetryable() && isRetryable(resp.Error(), c.wire, ctx) {
--- a/client.go
+++ b/client.go
@@ -272,7 +272,7 @@ func (c *dedicatedSingleClient) Do(ctx context.Context, cmd Completed) (resp Val
 	attempts := 1
 retry:
 	if err := c.check(); err != nil {
-		return newErrResult(err)
+		return NewErrorResult(err)
 	}
 	resp = c.wire.Do(ctx, cmd)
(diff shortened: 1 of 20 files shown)


rueian:
LGTM. @cfilby, just need you to sign off on your commit to make the DCO check pass.

cfilby:
> LGTM. @cfilby, just need you to sign off on your commit to make the DCO check pass.

Done! Followed the commands in the DCO Step.

Any other follow up steps I'd need to take? Guessing you or someone else would actually cut a new release at some point in the next week or two?

rueian (approved):

bradengroom:
Thank you @cfilby

TheDigitalBrian:
@cfilby helping out his Bham eng friends without even knowing ✊

PR #150, opened 2026-08-18
fix(pipe): avoid nil cache panic with disabled static-TTL multi-cache

kurochan:
## Summary

Fix a nil pointer panic when `DoMultiCache` is used with multiple
`ToStaticTTL()` commands while client-side caching is disabled.

## Root cause

With `DisableCache: true`, the pipe does not initialize `p.cache`.

The `DoMultiCache` fallback converted `Cacheable` commands to `Completed`
commands but preserved the internal static-TTL flag. In the background read
loop, the second and subsequent replies then entered the static-TTL path and
called `p.cache.Update` or `p.cache.Cancel`, causing a nil pointer panic.

## Changes

- Clear static-TTL metadata before forwarding disabled-cache requests to
  `DoMulti`.
- Guard the background static-TTL path when no cache store exists.
- Add regression tests for successful replies and Valkey errors with multiple
  static-TTL commands and a cancelable context.

## Compatibility

- `DisableCache` remains equivalent to `Do`/`DoMulti`.
- Client-side caching behavior is unchanged when enabled.
- Valkey errors are returned to callers unchanged.

### DIFF
--- a/pipe.go
+++ b/pipe.go
@@ -1713,7 +1713,9 @@ func (p *pipe) DoMultiCache(ctx context.Context, multi ...CacheableTTL) *valkeyr
 	if p.cache == nil {
 		commands := make([]Completed, len(multi))
 		for i, ct := range multi {
-			commands[i] = Completed(ct.Cmd)
+			cmd := ct.Cmd
+			cmds.ClearStaticTTL(&cmd)
+			commands[i] = Completed(cmd)
 		}
 		return p.DoMulti(ctx, commands...)
 	}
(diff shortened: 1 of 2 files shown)


pipe.go
```diff
@@ -647,7 +647,7 @@ func (p *pipe) _backgroundRead() (err error) {
 			if multi == nil {
 				multi = ones
 			}
-		} else if ff > 0 && cmds.IsStaticTTL(multi[ff]) {
+		} else if p.cache != nil && ff > 0 && cmds.IsStaticTTL(multi[ff]) {
```
rueian: Is it possible to avoid the nil check here?

kurochan: Thanks, addressed by removing the redundant nil check. https://github.com/valkey-io/valkey-go/pull/150/changes/2de1d221b5a1153605a793c50eb05fff3e8cbedf

rueian:
Hi @kurochan, please make the DCO check pass by signing off your commit.

kurochan:
@rueian Sorry! Fixed and signed off now.

rueian (approved):

PR #151, opened 2026-08-19
fix(cluster): re-pick unfinished commands between retry iterations

### LINKED ISSUE #59: Pipe commands getting stuck intermittently during scale up and scale down of elasticache clusters
We’re using valkey-go v1.0.57 to connect to an ElastiCache cluster with around 6 shards and 4 replicas, with replica auto-scaling enabled. During scale-up or scale-down events, we intermittently observe that `pipe.Exec(...)` calls hang indefinitely, despite having both connection timeouts and context deadlines set to 2 seconds.

We have faced this issue with pipelined HGetAll and Get commands.

Relevant client configurations:
```
SendToReplicas: true
Connection timeout: 2s
Ctx Deadline for each command: 2s
```
We’re trying to understand why this indefinite hang happens during resharding or topology changes.

DhruvilK7:
## Problem

Under `SendToReplicas: cmd.IsReadOnly()`, `context.Background()`, and `retry=true`, `clusterClient.DoMulti` and `DoMultiCache` re-bucketed `RedirectRetry` failures under the same conn key that produced them.

When a replica was removed from the cluster mid-flight:

- The retry envelope kept dispatching the same still-unfinished commands to the dead conn iteration after iteration.
- `lazyRefresh` fired but is async and coalesced, so the executing `DoMulti` never re-consulted the refreshed slot map.
- `WaitForRetry` sleeps up to 1s via a bare `time.Sleep` when `context.Background().Done() == nil`.

Result: unbounded retry loop, `err=0` returned to the caller.

## Fix

Between `WaitForRetry` and `goto retry`, walk `retries.m` and re-pick each unfinished non-ASK command against the current slot map via `c._pick(cmd.Slot(), c.toReplica(cmd))`. When `lazyRefresh` has converged the retry sees a fresh conn and reroutes.

- **ASK sub-buckets** (`nr.cAskings` / `nr.aIndexes`) stay pinned to the target picked by `redirectOrNew` on the previous iteration.
- **`MULTI`..`EXEC` transaction spans** move as a single unit to `_pick(spanSlot, false)` (primary, never replica), so a transaction never splits across two conns.
- **Emptied buckets** are returned to the `retryp` pool.
- **Cache path** mirrored in `resultcachefn` / `DoMultiCache` envelope with `rebucketRetriesCache` using `retrycachep`.

## Behaviour preserved

Behaviour is unchanged when `retries.m` contains only ASK entries or only conns still authoritative for their slots (`_pick` returns the same conn as the bucket key and the command stays put).

## Tests

Added to `cluster_test.go`:

- **`TestClusterDoMultiRepicksAfterReplicaRemoval`**: primary repro; hangs before the fix, passes after.
- **`TestClusterDoMultiPreservesAskingBucketing`**: guards that ASK bucket routing is not disturbed by the re-pick.
- **`TestClusterDoMultiCacheRepicksAfterReplicaRemoval`**: cache-path repro.
- **`TestClusterDoMultiTransactionStaysCoherent`**: `MULTI`..`EXEC` span moves as a unit, does not split across conns.

Fixes #59.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -941,9 +941,6 @@ func (c *clusterClient) doresultfn(
 // A MULTI..EXEC transaction span is moved as a unit to _pick(slot, false)
 // (the primary) so a transaction never splits across two conns.
 func (c *clusterClient) rebucketRetries(retries *connretry) {
-	if len(retries.m) == 0 {
-		return
-	}
 	type moved struct {
 		nc     conn
 		cIndex int
@@ -1458,9 +1455,6 @@ func (c *clusterClient) resultcachefn(
 // Cacheable commands never contain MULTI/EXEC so the transaction pinning
 // is omitted, but the ASK-preservation and empty-bucket cleanup match.
 func (c *clusterClient) rebucketRetriesCache(retries *connretrycache) {
-	if len(retries.m) == 0 {
-		return
-	}
 	type moved struct {
 		nc     conn
 		cIndex int
--- a/cluster.go
+++ b/cluster.go
@@ -957,46 +957,34 @@ func (c *clusterClient) rebucketRetries(retries *connretry) {
 		n := len(nr.commands)
 		keepIdx := nr.cIndexes[:0]
 		keepCmd := nr.commands[:0]
-		i := 0
-		for i < n {
-			// Locate the next MULTI or EXEC. Everything before it is a
-			// non-transaction run that can be re-picked per command.
-			txStart := n
-			for j := i; j < n; j++ {
-				if isMulti(nr.commands[j]) || isExec(nr.commands[j]) {
-					txStart = j
-					break
-				}
-			}
-			for j := i; j < txStart; j++ {
-				cmd := nr.commands[j]
+		for i := 0; i < n; {
+			if !isMulti(nr.commands[i]) {
+				cmd := nr.commands[i]
 				nc := c._pick(cmd.Slot(), c.toReplica(cmd))
 				if nc == nil || nc == oldConn {
-					keepIdx = append(keepIdx, nr.cIndexes[j])
+					keepIdx = append(keepIdx, nr.cIndexes[i])
 					keepCmd = append(keepCmd, cmd)
-					continue
+				} else {
+					moves = append(moves, moved{nc: nc, cIndex: nr.cIndexes[i], cmd: cmd})
 				}
-				moves = append(moves, moved{nc: nc, cIndex: nr.cIndexes[j], cmd: cmd})
+				i++
+				continue
 			}
-			if txStart == n {
-				break
+			// Find the matching EXEC. If the bucket only holds a partial
+			// span (which can happen if the original MULTI reply was not
+			// OK), no EXEC is found and the span runs to the end of the
+			// bucket instead; either way, keep the span intact.
+			j := i + 1
+			for j < n && !isExec(nr.commands[j]) {
+				j++
 			}
-			// Extend the transaction span through its matching EXEC. If the
-			// bucket only holds a partial span (which can happen if the
-			// original MULTI reply was not OK) span runs to the end of the
-			// bucket; either way, keep the span intact.
-			txEnd := txStart
-			for k := txStart; k < n; k++ {
-				if isExec(nr.commands[k]) {
-					txEnd = k
-					break
-				}
-				txEnd = k
+			if j == n {
+				j = n - 1
 			}
 			// Pick a slot for the whole span. MULTI/EXEC carry InitSlot, so
 			// prefer the slot of a real key inside the span.
-			spanSlot := nr.commands[txStart].Slot()
-			for k := txStart; k <= txEnd; k++ {
+			spanSlot := nr.commands[i].Slot()
+			for k := i; k <= j; k++ {
 				if s := nr.commands[k].Slot(); s != cmds.InitSlot {
 					spanSlot = s
 					break
@@ -1007,16 +995,16 @@ func (c *clusterClient) rebucketRetries(retries *connretry) {
 				nc = oldConn
 			}
 			if nc == oldConn {
-				for k := txStart; k <= txEnd; k++ {
+				for k := i; k <= j; k++ {
 					keepIdx = append(keepIdx, nr.cIndexes[k])
 					keepCmd = append(keepCmd, nr.commands[k])
 				}
 			} else {
-				for k := txStart; k <= txEnd; k++ {
+				for k := i; k <= j; k++ {
 					moves = append(moves, moved{nc: nc, cIndex: nr.cIndexes[k], cmd: nr.commands[k]})
 				}
 			}
-			i = txEnd + 1
+			i = j + 1
 		}
 		nr.cIndexes = keepIdx
 		nr.commands = keepCmd
--- a/cluster.go
+++ b/cluster.go
@@ -887,6 +887,108 @@ func (c *clusterClient) doresultfn(
 	return clean
 }
 
+// rebucketRetries re-picks the non-ASK portion of retries.m against the
+// current slot map. Called after WaitForRetry so lazyRefresh has had time to
+// converge on a live replica or the primary. ASK sub-buckets (cAskings /
+// aIndexes) stay pinned to the conn that redirectOrNew already selected.
+// A MULTI..EXEC transaction span is moved as a unit to _pick(slot, false)
+// (the primary) so a transaction never splits across two conns.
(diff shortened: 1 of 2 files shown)


rueian:
Hi @DhruvilK7, could you help resolve the conflicts?

DhruvilK7:
Hi @rueian, I resolved the conflicts

cluster.go
```diff
@@ -934,6 +934,108 @@ func (c *clusterClient) doresultfn(
...
+			txStart := n
+			for j := i; j < n; j++ {
+				if isMulti(nr.commands[j]) || isExec(nr.commands[j]) {
+					txStart = j
+					break
+				}
+			}
+			for j := i; j < txStart; j++ {
+				cmd := nr.commands[j]
+				nc := c._pick(cmd.Slot(), c.toReplica(cmd))
```
rueian, line 960: Hi @DhruvilK7, instead of having a new rebucketRetries function and doing the rebucketing right before we jump to the retry label, I think we should just use `c._pick` in the `doresultfn` under the `mode == RedirectRetry` block. That should simplify a lot. Same for the `rebucketRetriesCache`. We won't need these 2 functions anymore.

DhruvilK7: Thanks for the review, I looked into this in detail.

The issue is timing, not the picking itself. `shouldRefreshRetry` calls `c.lazyRefresh()`, which only schedules a refresh, it does not run one. `doresultfn` runs right when the error comes back, so calling `c._pick` there would just read the old, not yet refreshed slot map, the same stale conn we already have. The `WaitForRetry` sleep is what gives the scheduled refresh time to actually finish. `rebucketRetries` runs after that wait, right before we jump back to retry, so it reads a slot map that has had a chance to update. If we move the pick into `doresultfn`, we lose that ordering.

I tried this change and ran the existing tests. `TestClusterDoMultiRepicksAfterReplicaRemoval` and `TestClusterDoMultiCacheRepicksAfterReplicaRemoval` both fail, the replica ends up getting called twice instead of once, so it does not fix the retry within the same round.

There is a second problem. `doresultfn` picks one conn for the failing command and applies it to the whole `MULTI..EXEC` span. Right now that conn is always the one the transaction was already sent on, since `_pickMulti` always sends transactions to the primary, never a replica. 
If we pick using `c._pick(cm.Slot(), c.toReplica(cm))` instead, and the command that failed is a retryable read that is allowed to go to a replica, the whole transaction can get routed to a replica on retry. I wrote a test for this, it passes on the current code and fails once the inline pick is used, the transaction gets sent to the replica.

rueian: I see. That makes sense.

For this `for i < n {` loop, I found it hard to follow. Can we do the following instead?

For each `i`, do an if else statment:
* if `nr.commands[i]` is a MULTI, we find the next j where `nr.commands[j]` is an EXEC and re-pick the connection for the entire block. Then skip the block by doing `i = j + 1`.
* else, treat the `nr.commands[i]` as a normal command.

DhruvilK7: Sounds good, that reads much cleaner. 
Applied it, kept the "no matching EXEC, span runs to end of bucket" fallback for the partial-span case, and added a test that locks in the replica-routing issue from my last reply as a regression check.

cluster.go
```diff
@@ -934,6 +934,96 @@ func (c *clusterClient) doresultfn(
...
+	for oldConn, nr := range retries.m {
+		if len(nr.commands) == 0 {
+			continue
+		}
+		n := len(nr.commands)
+		keepIdx := nr.cIndexes[:0]
+		keepCmd := nr.commands[:0]
+		for i := 0; i < n; {
+			if !isMulti(nr.commands[i]) {
+				cmd := nr.commands[i]
```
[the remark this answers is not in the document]
rueian: @DhruvilK7, this review looks valid. Is repicking a connection for `slot == cmds.InitSlot` randomly here intentional?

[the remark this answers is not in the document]
DhruvilK7: Not intentional as a deliberate choice for this path, but I don't think CodeRabbit's fix is actually safe, so I'd rather leave it.

`_pick(InitSlot, ...)` only ever returns something from `c.conns`, and that map only ever holds the current live set of conns, a removed node gets deleted from it. So the "random" pick can shuffle a keyless command to a different live node, but never to a dead one.

I tried CodeRabbit's exact diff and ran it through an actual `DoMulti` retry. If the conn a keyless command is sitting on gets dropped from the topology mid-retry, this fix pins it there for good: 3.8 million retries against the dead conn in 2 seconds, 0 reaching a live one, never succeeds. Same failure mode as the original issue, just for keyless commands.

I also tried gating the skip on whether `oldConn` is still in `c.conns`, that one is correct, but it adds a map read under lock per keyless command per retry round for a problem that's cosmetic, not a bug.

So I think, current behavior is safe, just not deterministic in an already rare case.

rueian:
> ## [Codecov](https://app.codecov.io/gh/valkey-io/valkey-go/pull/151?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=valkey-io) Report
> ❌ Patch coverage is `82.60870%` with `16 lines` in your changes missing coverage. Please review. ✅ Project coverage is 94.00%. Comparing base ([`32e48cc`](https://app.codecov.io/gh/valkey-io/valkey-go/commit/32e48cc2adcbdba7209798208176ba3f44460bfd?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=valkey-io)) to head ([`8047a73`](https://app.codecov.io/gh/valkey-io/valkey-go/commit/8047a73dafed06ba1d364bfc3b7cc34b45a2dd7b?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=valkey-io)). ⚠️ Report is 4 commits behind head on main.
> 
> [Files with missing lines](https://app.codecov.io/gh/valkey-io/valkey-go/pull/151?dropdown=coverage&src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=valkey-io)	Patch %	Lines
> [cluster.go](https://app.codecov.io/gh/valkey-io/valkey-go/pull/151?src=pr&el=tree&filepath=cluster.go&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=valkey-io#diff-Y2x1c3Rlci5nbw==)	82.60%	[9 Missing and 7 partials ⚠️ ](https://app.codecov.io/gh/valkey-io/valkey-go/pull/151?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=valkey-io)
> Additional details and impacted files
> ```diff
> @@            Coverage Diff             @@
> ##             main     #151      +/-   ##
> ==========================================
> + Coverage   93.45%   94.00%   +0.55%     
> ==========================================
>   Files         107      107              
>   Lines       52626    49511    -3115     
> ==========================================
> - Hits        49180    46545    -2635     
> + Misses       2938     2450     -488     
> - Partials      508      516       +8     
> ```
> 
> [☔ View full report in Codecov by Harness](https://app.codecov.io/gh/valkey-io/valkey-go/pull/151?dropdown=coverage&src=pr&el=continue&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=valkey-io). 📢 Have feedback on the report? [Share it here](https://about.codecov.io/codecov-pr-comment-feedback/?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=valkey-io).
> 
> 🚀 New features to boost your workflow:

@DhruvilK7, LGTM now, but could you try to make the coverage 100%? I also feel that some early returns may not be necessary, like the one with the `len(retries.m) == 0` check.

DhruvilK7:
@rueian have tried to make coverage 100% in the latest commit and also removed early returns as suggested.

rueian (approved):

rueian:
Thanks @DhruvilK7!

cluster.go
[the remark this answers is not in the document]
rueian: > I also tried gating the skip on whether oldConn is still in c.conns, that one is correct, but it adds a map read under lock per keyless command per retry round for a problem that's cosmetic, not a bug.

@DhruvilK7, I think it is still worth trying to reduce the number of conns we use. How about we just pick the previous conn we repicked during the loop? We can try this in a follow-up PR.

Issue #153, opened 2026-08-21
Cluster client never routes reads to newly-added replicas when ShardsRefreshInterval is unset (default 0)


DhruvilK7:
With `ClusterOption.ShardsRefreshInterval` unset (default `0`), the cluster client never re-scans topology unless a command errors with `MOVED/ASK/CLUSTERDOWN/TRYAGAIN/LOADING`, or a connection closes. Scaling up replica count on an existing shard (e.g. AWS ElastiCache/Valkey "modify shard configuration") does neither — no slot movement, no error on existing conns.

## Suggested fix
Give `ShardsRefreshInterval` a sane non-zero default (e.g. 10s same as go-redis) instead of "disabled"

rueian:
Hi @DhruvilK7, valkey-go won't read replicas by default, so we are not going to enable ShardsRefreshInterval by default as well. But we can improve the README by adding ShardsRefreshInterval alongside where SendToReplicas is used, with some explanation in a comment.

PR #154, opened 2026-08-23
docs: mention ShardsRefreshInterval alongside SendToReplicas for cluster replicas

DhruvilK7:
This is to address: #153

### DIFF
--- a/README.md
+++ b/README.md
@@ -501,8 +501,20 @@ client, err := valkey.NewClient(valkey.ClientOption{
     SendToReplicas: func(cmd valkey.Completed) bool {
         return cmd.IsReadOnly()
     },
+    ClusterOption: valkey.ClusterOption{
+        // Set this to periodically refresh the cluster topology.
+        // Without it, newly added or removed replicas won't be picked up,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -501,8 +501,20 @@ client, err := valkey.NewClient(valkey.ClientOption{
...
+        // Without it, newly added or removed replicas won't be picked up,
+        // because the topology is only refreshed on MOVED/ASK errors or connection closures.
+        ShardsRefreshInterval: 10 * time.Second,
+    },
 })
+```
 
+Note that when you opt into reading from replicas on a cluster, scaling the number of replicas on an existing shard
+(e.g. via ElastiCache modify-replication-group) neither moves slots nor closes existing connections,
+so without `ShardsRefreshInterval` the client never discovers the new replicas.
```
rueian:
```suggestion
Note that scaling the number of replicas on an existing shard (e.g. via ElastiCache modify-replication-group)
neither moves slots nor closes existing connections, so without `ShardsRefreshInterval`
the client never discovers the new replicas and continues reading the old shard.
```

rueian (commented):
LGTM. Thanks!

PR #155, opened 2026-08-28
gap : added get to buffer gap

ankitg1706:
file cahnges for get to buffer


valkeycompat/adapter.go
```diff
@@ -1003,6 +1003,28 @@ func (c *Compat) Get(ctx context.Context, key string) *StringCmd {
 	resp := c.client.Do(ctx, cmd)
 	return newStringCmd(resp)
 }
+func (c *Compat) GetToBuffer(ctx context.Context, key string, buf []byte) *ZeroCopyStringCmd {
```
rueian, line 1006:
```suggestion

func (c *Compat) GetToBuffer(ctx context.Context, key string, buf []byte) *ZeroCopyStringCmd {
```

valkeycompat/adapter.go
```diff
@@ -1003,6 +1003,28 @@ func (c *Compat) Get(ctx context.Context, key string) *StringCmd {
...
+	cmd := &ZeroCopyStringCmd{
+		buf: buf,
+	}
+
+	resp := c.client.Do(ctx, c.client.B().Get().Key(key).Build())
```
rueian, line 1011: Shouldn't we use DoStream for being zero copy?

valkeycompat/getToBuffer_test.go
```diff
@@ -0,0 +1,252 @@
+package valkeycompat
```
rueian, line 1: can we merge this file into adapter_test.go?

valkeycompat/pipeline.go
```diff
@@ -367,6 +367,11 @@ func (c *Pipeline) Get(ctx context.Context, key string) *StringCmd {
 	c.rets = append(c.rets, ret)
 	return ret
 }
+func (c *Pipeline) GetToBuffer(ctx context.Context, key string, buf []byte) *ZeroCopyStringCmd {
+	ret := c.comp.GetToBuffer(ctx, key, buf)
```
rueian, line 371: If we make GetToBuffer zero copy, then it won't support Pipeline and TxPipeline. We can just panic here.

rueian:
duplicated

PR #159, opened 2026-08-31
feat: add RouteByLatency option for sentinel replica selection

kelle-konstantin:
### Summary

This PR adds the `RouteByLatency` option to `ClientOption` for Sentinel-managed clients, allowing read traffic to be routed to the replica with the lowest round-trip latency (e.g. co-located local nodes or same-rack/AZ replicas).

### Motivation

Currently, when `SendToReplicas` or `ReplicaOnly` is enabled in Sentinel mode, `valkey-go` selects an eligible replica uniformly at random (`util.FastRand`).

In many Kubernetes and multi-node deployments, Valkey replicas are deployed alongside application pods (e.g. as a DaemonSet or per-node instance). In such topologies:
- Reading from a co-located local replica on the same node achieves sub-millisecond latency (<0.1ms) with zero cross-node network overhead.
- Selecting a replica at random frequently routes requests across nodes or AZs, incurring unnecessary 1–2ms round-trip latency and network bandwidth.

Similar to `go-redis`'s `RouteByLatency`, this PR enables clients to automatically discover and prefer the lowest-latency replica while preserving full failover and failback capabilities.

---

### Key Changes

1. **`ClientOption.RouteByLatency`**:
   - Added a new boolean field `RouteByLatency` to `ClientOption`.
   - When set to `true`:
     - Automatically routes read commands to replicas (`SendToReplicas = func(cmd Completed) bool { return true }`) if not explicitly specified.
     - Defaults `Sentinel.TopologyRefreshInterval` to `10s` if unset, enabling seamless periodic reconciliation and failback.

2. **Concurrent Latency Probing (`pickLowestLatencyReplica`)**:
   - Probes all non-`sdown` replicas in parallel using lightweight goroutines.
   - Measures end-to-end RTT using `dialer.DialContext` followed by a RESP `PING` (`*1\r\n$4\r\nPING\r\n`) to verify both TCP reachability and engine responsiveness.
   - Respects `dialer.Timeout` (or a 500ms fallback) to prevent blocking on hung nodes.

3. **Connection Stability & Failback Support**:
   - **Anti-Churn / Hysteresis Threshold (200µs)**: If the client is already connected to a healthy replica and its latency is within 200µs of the fastest node, the existing connection is retained to prevent socket reconnection churn during periodic refresh ticks.
   - **Automatic Failback**: If the local replica temporarily goes down and later recovers, the periodic refresh reconciler detects the lower latency and seamlessly migrates back to the local replica via atomic connection swap (`_switchTarget`).

4. **Zero Hot-Path Overhead**:
   - Latency probing only runs during initial connection or topology reconciliation. Regular command execution (`Do`, `DoMulti`, `DoCache`) continues to use atomic connection loading with zero added allocations.

---

### Test Coverage

Added comprehensive unit tests in `route_by_latency_test.go`:
- `TestPickLowestLatencyReplica`: Verifies that the lowest-latency node is consistently chosen among replicas with simulated network delays (0ms vs 30ms vs 60ms).
- `TestPickLowestLatencyReplica_FailbackToFasterNode`: Verifies that a client connected to a remote replica automatically migrates back when a lower-latency node recovers.
- `TestPickLowestLatencyReplica_WithUnreachableNodes`: Verifies graceful fallback when some nodes are offline or unreachable.
- `TestSentinelAutoEnableSendToReplicasAndRefreshInterval`: Verifies proper default configuration when `RouteByLatency` is enabled.

```bash
$ go test -v -run 'TestPickLowestLatencyReplica|TestSentinelAutoEnableSendToReplicasAndRefreshInterval' .
=== RUN   TestPickLowestLatencyReplica
--- PASS: TestPickLowestLatencyReplica (0.31s)
=== RUN   TestPickLowestLatencyReplica_FailbackToFasterNode
--- PASS: TestPickLowestLatencyReplica_FailbackToFasterNode (0.05s)
=== RUN   TestPickLowestLatencyReplica_WithUnreachableNodes
--- PASS: TestPickLowestLatencyReplica_WithUnreachableNodes (0.00s)
=== RUN   TestSentinelAutoEnableSendToReplicasAndRefreshInterval
--- PASS: TestSentinelAutoEnableSendToReplicasAndRefreshInterval (0.00s)
PASS
ok      github.com/valkey-io/valkey-go  0.848s

### DIFF
--- a/valkey.go
+++ b/valkey.go
@@ -269,6 +269,10 @@ type ClientOption struct {
 	//  if valkey instance is a cluster or a single valkey instance.
 	ForceSingleClient bool
 
+	// RouteByLatency when connecting to Sentinel with SendToReplicas (or ReplicaOnly),
+	// measures network/PING latency to all eligible replicas and selects the one with the lowest latency.
+	RouteByLatency bool
+
 	// ReplicaOnly indicates that this client will only try to connect to readonly replicas of valkey setup.
 	ReplicaOnly bool
 
--- a/valkey.go
+++ b/valkey.go
@@ -271,6 +271,9 @@ type ClientOption struct {
 
 	// RouteByLatency when connecting to Sentinel with SendToReplicas (or ReplicaOnly),
 	// measures network/PING latency to all eligible replicas and selects the one with the lowest latency.
+	// If SendToReplicas is nil and ReplicaOnly is false, the sentinel client sets SendToReplicas
+	// automatically so that read commands can reach replicas.
+	// This option has no effect on the cluster, standalone, or single clients.
 	RouteByLatency bool
 
 	// ReplicaOnly indicates that this client will only try to connect to readonly replicas of valkey setup.
--- a/route_by_latency_test.go
+++ b/route_by_latency_test.go
@@ -1,8 +1,10 @@
 package valkey
 
 import (
+	"context"
 	"fmt"
 	"net"
+	"strconv"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: valkey.go [hunk 1]
```diff
@@ -269,6 +269,13 @@ type ClientOption struct {
 	//  if valkey instance is a cluster or a single valkey instance.
 	ForceSingleClient bool
 
+	// RouteByLatency when connecting to Sentinel with SendToReplicas (or ReplicaOnly),
+	// measures network/PING latency to all eligible replicas and selects the one with the lowest latency.
+	// If SendToReplicas is nil and ReplicaOnly is false, the sentinel client sets SendToReplicas
+	// automatically so that read commands can reach replicas.
+	// This option has no effect on the cluster, standalone, or single clients.
+	RouteByLatency bool
```
rueian, line 277: Hi @kelle-konstantin, we probably should not add this field. We should try to reuse the ReadNodeSelectorFunc. For example, make the replica selection call ReadNodeSelectorFunc so that users can do the RouteByLatency themselves.

PR #161, opened 2026-09-01
fix(cluster): close conns added to c.conns while _refresh is unlocked

### LINKED ISSUE #160: Cluster client leaks a connection when redirectOrNew races with _refresh
`clusterClient._refresh` reads `c.conns` under `RLock` and builds `removes` there, releases the lock, computes the slot table, and only then overwrites `c.conns` under `Lock`.

https://github.com/valkey-io/valkey-go/blob/v1.0.77/cluster.go#L260-L271

https://github.com/valkey-io/valkey-go/blob/v1.0.77/cluster.go#L343-L347

On a MOVED reply, `do()` schedules `_refresh` through `lazyRefresh()` (random delay of up to 1s) and then calls `redirectOrNew()`, which inserts a new mux into `c.conns`. The following `ncc.Do` establishes a connection on it.

https://github.com/valkey-io/valkey-go/blob/v1.0.77/cluster.go#L626-L632

If that insertion lands after `_refresh` released the read lock, the new conn is:

- absent from the local `conns` map, so `c.conns = conns` drops it
- absent from `removes`, which was computed before the insertion, so it is never closed
- unreachable from `c.conns`, so it never receives the `InitSlot` commands (e.g. `PING`) that `_pick` spreads over `c.conns` and that would otherwise fill its remaining wires

The mux keeps its background goroutines, so it is not garbage collected either. The connection stays established and idle until the process exits.

The `prev == cc.conn` branch of `redirectOrNew` has the same problem in reverse: the overwrite reinstates the old conn and drops the replacement without closing it.

## Reproduction

With no changes to the library, 64 goroutines issuing `GET` against a mocked cluster that answers MOVED with a fresh address each time leaked 21 conns in 40s (`created=1421 closed=1400 liveConns=0`, counting `connFn`/`Close()` calls and entries left in `c.conns`).

## Impact

We hit this against ElastiCache Serverless (valkey 8.1), which relocates slots internally and so returns MOVED occasionally. Individual clients went from 8 to 9 connections and stayed there: `valkey_dial_success` never diverged from `valkey_dial_conns`, so nothing was ever closed, and the extra conn gained no further wires despite steady `PING` traffic. The server side recorded exactly one new connection at the same minute.

Bounded per event, but unbounded over time on workloads where MOVED is routine.

`redis/rueidis` carries the same code at the same lines, so it is presumably affected too.

naoto-naka-safie:
## Problem

`_refresh` reads `c.conns` under `RLock` and builds `removes` there, releases the lock, computes `wslots`, and only then overwrites `c.conns` under `Lock`. A conn that `redirectOrNew` inserts into `c.conns` during that window is dropped by the overwrite without being closed, so its connection stays established for the lifetime of the process.

Fixes #160

## Fix

Build the removal set under the write lock, comparing against the `conns` map that is about to be installed. An entry is closed when it is absent from that map, or when its conn differs from the one being installed.

The second condition also covers the `prev == cc.conn` branch of `redirectOrNew`, where the addr is unchanged but the conn was replaced: previously the overwrite reinstated the old conn and dropped the replacement without closing it.

## Tests

`TestClusterClientRefreshClosesConnAddedWhileUnlocked` reproduces the window deterministically, with no timing dependence. With `EnableReplicaAZInfo` set, `_refresh` calls `conn.AZ()` after releasing the read lock and before taking the write lock, so the mock's `AZFn` invokes `redirectOrNew` at exactly that point.

The test fails before this change (the conn is dropped from `c.conns` and never closed) and passes after it. `go test -race` on the cluster tests reports no data race.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -360,16 +360,28 @@ func (c *clusterClient) _refresh() (err error) {
 			conns[addr] = fresh
 		}
 	}
-	if replaced != nil {
-		for i := range wslots {
-			if cc, ok := replaced[wslots[i]]; ok {
-				wslots[i] = cc
+	for _, g := range groups {
+		hit := false
+		for i := range g.nodes {
+			if _, ok := replaced[g.nodes[i].conn]; ok {
+				hit = true
+				break
 			}
 		}
-		for _, nodes := range rslots {
-			for i := range nodes {
-				if cc, ok := replaced[nodes[i].conn]; ok {
-					nodes[i].conn = cc
+		if !hit {
+			continue
+		}
+		for _, slot := range g.slots {
+			for i := slot[0]; i <= slot[1] && i >= 0 && i < 16384; i++ {
+				if cc, ok := replaced[wslots[i]]; ok {
+					wslots[i] = cc
+				}
+				if rslots != nil {
+					for j := range rslots[i] {
+						if cc, ok := replaced[rslots[i][j].conn]; ok {
+							rslots[i][j].conn = cc
+						}
+					}
 				}
 			}
 		}
--- a/cluster.go
+++ b/cluster.go
@@ -360,26 +360,28 @@ func (c *clusterClient) _refresh() (err error) {
 			conns[addr] = fresh
 		}
 	}
-	for _, g := range groups {
-		hit := false
-		for i := range g.nodes {
-			if _, ok := replaced[g.nodes[i].conn]; ok {
-				hit = true
-				break
-			}
-		}
-		if !hit {
-			continue
-		}
-		for _, slot := range g.slots {
-			for i := slot[0]; i <= slot[1] && i >= 0 && i < 16384; i++ {
-				if cc, ok := replaced[wslots[i]]; ok {
-					wslots[i] = cc
+	if replaced != nil {
+		for _, g := range groups {
+			hit := false
+			for i := range g.nodes {
+				if _, ok := replaced[g.nodes[i].conn]; ok {
+					hit = true
+					break
 				}
-				if rslots != nil {
-					for j := range rslots[i] {
-						if cc, ok := replaced[rslots[i][j].conn]; ok {
-							rslots[i][j].conn = cc
+			}
+			if !hit {
+				continue
+			}
+			for _, slot := range g.slots {
+				for i := slot[0]; i <= slot[1] && i >= 0 && i < 16384; i++ {
+					if cc, ok := replaced[wslots[i]]; ok {
+						wslots[i] = cc
+					}
+					if rslots != nil {
+						for j := range rslots[i] {
+							if cc, ok := replaced[rslots[i][j].conn]; ok {
+								rslots[i][j].conn = cc
+							}
 						}
 					}
 				}
--- a/cluster.go
+++ b/cluster.go
@@ -257,15 +257,11 @@ func (c *clusterClient) _refresh() (err error) {
 		}
 	}
 
-	var removes []conn
-
 	c.mu.RLock()
 	for addr, cc := range c.conns {
 		if fresh, ok := conns[addr]; ok {
 			fresh.conn = cc.conn
 			conns[addr] = fresh
-		} else {
-			removes = append(removes, cc.conn)
 		}
 	}
 	c.mu.RUnlock()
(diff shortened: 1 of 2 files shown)


cluster.go
```diff
@@ -340,7 +336,16 @@ func (c *clusterClient) _refresh() (err error) {
...
 	}
 
+	var removes []conn
+
 	c.mu.Lock()
+	// redirectOrNew may have changed c.conns while the lock was released above,
+	// so collect the removals here. Otherwise those conns are dropped without being closed.
+	for addr, cc := range c.conns {
+		if fresh, ok := conns[addr]; !ok || fresh.conn != cc.conn {
+			removes = append(removes, cc.conn)
```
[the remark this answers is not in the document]
naoto-naka-safie: Good catch, fixed in 39e731b.

When the conn at an unchanged addr differs, `_refresh` now keeps the replacement instead of reinstating the conn `redirectOrNew` already scheduled for closing, and retargets `wslots`/`rslots` at it. The retarget scan only runs when a replacement was actually observed.

Added `TestClusterClientRefreshKeepsConnReplacedWhileUnlocked`, which replaces the conn at an unchanged addr from inside the unlocked window and asserts that both `c.conns` and `wslots` end up pointing at the replacement. It fails without this commit.

cluster.go
```diff
@@ -340,7 +336,43 @@ func (c *clusterClient) _refresh() (err error) {
...
+			continue
+		}
+		if fresh.conn != cc.conn {
+			// redirectOrNew replaced the conn at this addr and already scheduled the
+			// one it replaced for closing, so keep the replacement and retarget the slots.
+			if replaced == nil {
+				replaced = make(map[conn]conn, 1)
+			}
+			replaced[fresh.conn] = cc.conn
+			fresh.conn = cc.conn
```
rueian, line 359: We will need to close the `fresh.conn` if the connection is already connected by the `info.AZ = info.conn.AZ()` above.
We can append `fresh.conn` to `removes` if we find that.

cluster.go
```diff
@@ -340,7 +336,43 @@ func (c *clusterClient) _refresh() (err error) {
...
+			conns[addr] = fresh
+		}
+	}
+	if replaced != nil {
+		for i := range wslots {
```
rueian: instead of looping each slots, is it possible to loop `groups` instead?

cluster.go
naoto-naka-safie: Good catch, fixed in 75cf8db — `fresh.conn` now goes into `removes` as well.

Added `TestClusterClientRefreshClosesConnConnectedByAZ` for it: a node appears only in the refreshed topology, so `_refresh` carries its own conn for it and `AZ()` connects it, and `redirectOrNew` then installs another conn at that addr from inside that `AZ()` call. It fails without this commit.

cluster.go
naoto-naka-safie: Looping `groups` by itself would not reach them: `wslots`/`rslots` already hold the conn values copied out of `g.nodes` earlier, so rewriting `g.nodes[i].conn` afterwards has no effect on the slot tables.

Two ways I can see to avoid the full 16384 scan:

1. Walk `groups`, and for a group whose node was replaced, rewrite only the slot ranges listed in `g.slots`. Narrower than the full scan, but more code.
2. Move the `wslots`/`rslots` construction inside the write lock so it is built once from the reconciled `conns`. Simpler, but it holds the write lock across the slot loop (the `AZ()` calls would stay outside it).

Which do you prefer? Happy to do either.

rueian: should be 1.

naoto-naka-safie: Done in 7180fdf. `_refresh` now walks `groups` and rewrites only the slot ranges of the ones whose node was replaced, instead of scanning all 16384 entries.

With that, the `replaced != nil` guard is no longer needed — every group short-circuits on the `hit` check when nothing was replaced. `rslots` still needs its nil check since it stays nil unless `SendToReplicas` is set.

cluster.go
```diff
@@ -340,7 +336,56 @@ func (c *clusterClient) _refresh() (err error) {
...
+			if replaced == nil {
+				replaced = make(map[conn]conn, 1)
+			}
+			replaced[fresh.conn] = cc.conn
+			removes = append(removes, fresh.conn)
+			fresh.conn = cc.conn
+			conns[addr] = fresh
+		}
+	}
+	for _, g := range groups {
```
rueian: if replaced == nil, we should skip the entire loop.

naoto-naka-safie: Done in 329cce6.

While fixing this I realised the retargeting only exists because the slot tables are built from `g.nodes[i].conn` before the conns are final. If that build moved after the reconciliation under the write lock, `replaced` and this whole loop could go away:

- `AZ()` returns a per-node string, so it stays valid across a conn swap and can still run before the lock, as it does today.
- The reconciliation would settle `conns` first, and `g.nodes[i].conn` plus `wslots`/`rslots` would then be built once from the final values.

The cost is holding the write lock across the slot loop — no I/O in there, `AZ()` stays outside it.

Would you prefer that shape, or keep the current one? I am happy to leave it as is if the lock duration matters more here.

rueian: Let's keep the current one for now. Holding the write lock all the time does simplify the code, but I am not sure if it is worth making such a big change. But you are welcome to investigate that later.

rueian (approved):

PR #162, opened 2026-09-01
Infomap

vviroje:
Info is a Valkey command that returns comprehensive server information, performance metrics, and usage statistics in an easy-to-read text format. But **InfoMap** is a wrapper that gives data in map.The compatibility gap is the absence of a dedicated InfoMap API in valkeycompat.
The implementation adds a compatibility-layer API while sending the native Info command to the Valkey server.


vviroje:
wrong PR

PR #163, opened 2026-09-02
Add InfoMap compatibility in Valkey-go

kumariam-ai:
**InfoMap** is a wrappe of Info Valkey command that returns comprehensive server information, performance metrics, and usage statistics in an easy-to-read map format.
The compatibility gap is the absence of a dedicated InfoMap API in valkeycompat.

**Purpose**: It queries the Valkey server using the INFO command and returns the statistics or server properties as a parsed map instead of a raw unformatted string.
**Usage**: You call it from a connected Valkey client instance by passing an optional section name (like "cpu", "memory", or "stats").

**### Testcases** :

**valkeycompat/adapter_test.go | It("should InfoMap", func()**
Adds an integration test for Adapter.InfoMap(), which sends the INFO command to the Valkey server and parses the raw multi-section text response into a structured map[string]map[string]string.

**valkeycompat/pipeline_test.go | p.InfoMap(ctx)**
Adds InfoMap to the TestPipeliner coverage test, which verifies that every Pipeliner method correctly buffers its command into the pipeline without executing it immediately.

**command_test.go | It("InfoMap", ...), It("parseInfoString", ...)**
Smoke test inside testCmd — a function that tests that the Adapter correctly bridges Go types ↔ Valkey wire protocol.
This test validates ValkeyResult.AsString, which parses an INFO-like multi-section text blob into map[string]map[string]string.

**### Logs:**
infomap

Signed-off-by: Amrita Kumari <kumariam@google.com>
Co-authored-by: Omkar Mestry <omanges@google.com>

### DIFF
--- a/message.go
+++ b/message.go
@@ -498,14 +498,6 @@ func (r ValkeyResult) AsIntMap() (v map[string]int64, err error) {
 	return
 }
 
-// AsStrStrMap delegates to ValkeyMessage.AsStrStrMap
-func (r ValkeyResult) AsStrStrMap() (map[string]map[string]string, error) {
-	if r.err != nil {
-		return nil, r.err
-	}
-	return r.val.AsStrStrMap()
-}
-
 // AsScanEntry delegates to ValkeyMessage.AsScanEntry.
 func (r ValkeyResult) AsScanEntry() (v ScanEntry, err error) {
 	if r.err != nil {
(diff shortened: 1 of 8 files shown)


message.go
```diff
@@ -1273,6 +1281,116 @@ func (m *ValkeyMessage) AsIntMap() (map[string]int64, error) {
 	return nil, fmt.Errorf("%w: valkey message type %s is not a map/array/set or its length is not even", errParse, typeNames[typ])
 }
 
+// AsStrStrMap parses INFO-like text into map[section]map[key]value.
+// Lines starting with '#' define sections; top-of-file keys go to "default".
+func (m *ValkeyMessage) AsStrStrMap() (map[string]map[string]string, error) {
```
rueian: Hi @kumariam-ai, mostly LGTM but this parsing function shouldn't be named as `AsStrStrMap` here because this parsing is specifically for the out of INFO command. Can we move this function to valkeycompat for now and keep it private? Also give it a name like parseInfoString.

kumariam-ai: Updated as suggested! I have removed AsStrStrMap from message.go / ValkeyMessage and moved the parsing logic into valkeycompat as a private function named parseInfoString. Thanks for the review!

rueian (approved):

PR #164, opened 2026-09-02
add MSetEx compatibility

sonalijhala-23:
## Summary
Adds `Compat.MSetEX` and `Pipeline.MSetEX` to the `valkeycompat` package to support the `MSETEX` command (introduced in Valkey / Redis 8.4+). This enables clients using the compatibility adapter to atomically set multiple key-value pairs with expiration options and existence conditions.

## Rationale
Valkey and Redis 8.4+ introduce `MSETEX` to allow setting multiple keys simultaneously with shared expiration semantics and conditional flags (`NX` / `XX`). Adding this to `valkeycompat` keeps the compatibility layer feature-complete and aligns with standard go-redis/rueidiscompat patterns without requiring low-level custom command construction.

---

### 1. Types & Definitions (`valkeycompat/command.go`)
- Added `ErrLocalValidation` for local client-side validation failures.
- Added `SetCondition` (`NX`, `XX`).
- Added `ExpirationMode` (`EX`, `PX`, `EXAT`, `PXAT`, `KEEPTTL`) and `ExpirationOption`.
- Added `MSetEXArgs` struct to encapsulate conditions and expiration settings.

### 2. Adapter Implementation (`valkeycompat/adapter.go`)
- Added `MSetEX(ctx context.Context, args MSetEXArgs, values ...any) *IntCmd` to `CoreCmdable` interface and `Compat`.
- Validates key/value pairs locally (requires non-empty, even number of key/value arguments, and valid expiration modes).
- Correctly registers keys using `.Keys()` for proper cluster routing, followed by `.Args()` for values, condition, and expiration parameters.
- Returns `*IntCmd` (returns `1` on success, `0` if condition was not met).

### 3. Pipeline Support (`valkeycompat/pipeline.go`)
- Added `Pipeline.MSetEX` proxy wrapper.
- Skips queuing into return slice (`c.rets`) if local argument validation fails (`ErrLocalValidation`), maintaining error consistency with other pipeline commands.

### 4. Tests & Golden File Updates
- **`valkeycompat/adapter_test.go`**: Comprehensive Ginkgo integration tests covering all expiration modes, existence conditions, edge cases, and argument validation across both RESP2 and RESP3.
- **`valkeycompat/pipeline_test.go`**: Updated command matrix golden expectations (incremented count from 491 to 492) and verified token sequence.
- **`valkeycompat/command_test.go`**: Added unit test verifying `Pipeline.MSetEX` correctly constructs the expected token sequence (`["MSETEX", "1", "k1", "v1", "EX", "1"]`).

---

## Files Changed

| File | Description |
| :--- | :--- |
| `valkeycompat/command.go` | Added `MSetEXArgs`, `ExpirationOption`, `SetCondition`, `ExpirationMode`, and `ErrLocalValidation` |
| `valkeycompat/adapter.go` | Added `CoreCmdable.MSetEX` interface definition and `Compat.MSetEX` implementation |
| `valkeycompat/pipeline.go` | Added `Pipeline.MSetEX` wrapper and local validation error handling |
| `valkeycompat/adapter_test.go` | Added test specs covering validation, expiration modes, conditions, and pipeline execution |
| `valkeycompat/pipeline_test.go` | Added `MSetEX` entry to pipeline golden expectation matrix (491 → 492) |
| `valkeycompat/command_test.go` | Added unit test ensuring pipeline builds the expected token sequence |

---

### Argument & Input Validation
- [x] **Empty values**: Rejects empty arguments with `ErrLocalValidation`.
- [x] **Empty map**: Rejects empty map arguments with `ErrLocalValidation`.
- [x] **Odd argument count**: Rejects odd number of key/value arguments and returns error-wrapped `IntCmd`.
- [x] **Invalid expiration mode**: Rejects unsupported expiration modes.

### Return Semantics
- [x] Returns `1` when all keys are successfully set.
- [x] Returns `0` when condition (`NX` / `XX`) prevents setting the keys (no-op).

### Expiration Modes
- [x] **`EX` (seconds)**: Confirms TTL is set within the specified seconds range.
- [x] **`PX` (milliseconds)**: Confirms PTTL behavior matches millisecond precision.
- [x] **`EXAT` (Unix timestamp in seconds)**: Validates absolute expiration timestamp.
- [x] **`PXAT` (Unix timestamp in milliseconds)**: Validates absolute millisecond timestamp.
- [x] **`KEEPTTL`**: Preserves pre-existing TTL on modified keys.

### Conditions & Multi-Key
- [x] **`NX`**: Only sets keys if none of them exist.
- [x] **`XX`**: Only sets keys if all already exist; no-ops if any key is missing.
- [x] **Multi-key (>2)**: Sets 3+ key/value pairs atomically with shared TTL.

### Pipeline Integration
- [x] `Pipeline.MSetEX` appends `MSETEX` command with exact tokens (`numkeys`, `key`, `val`, condition, expiration).
- [x] Verified in pipeline golden test matrix (492 calls).
- [x] Both RESP2 and RESP3 protocols pass all specs.

---

## Verification & Test Logs

<details>
<summary><b>1. Ginkgo Adapter Suite: <code>go test -v -run TestAdapter -ginkgo.focus="MSetEX"</code> (Click to expand)</b></summary>

```text
=== RUN   TestAdapter
Running Suite: Adapter Suite - /usr/local/google/home/sonalijhala/Valkey-Repo/valkey-go/valkeycompat
Random Seed: 1788370393

Will run 22 of 889 specs
valkeycompat PTTL compare1 = 10000 ms
valkeycompat PTTL compare2 = 10000 ms
valkeycompat PTTL compare1 = 10000 ms
valkeycompat PTTL compare2 = 10000 ms

Ran 22 of 889 Specs in 20.717 seconds
SUCCESS! -- 22 Passed | 0 Failed | 0 Pending | 867 Skipped
--- PASS: TestAdapter (20.74s)
PASS
ok  	github.com/valkey-io/valkey-go/valkeycompat	20.812s

### DIFF
--- a/valkeycompat/adapter.go
+++ b/valkeycompat/adapter.go
@@ -125,6 +125,7 @@ type CoreCmdable interface {
 	MGet(ctx context.Context, keys ...string) *SliceCmd
 	MSet(ctx context.Context, values ...any) *StatusCmd
 	MSetNX(ctx context.Context, values ...any) *BoolCmd
+	MSetEX(ctx context.Context, args MSetEXArgs, values ...any) *IntCmd
 	Set(ctx context.Context, key string, value any, expiration time.Duration) *StatusCmd
 	SetArgs(ctx context.Context, key string, value any, a SetArgs) *StatusCmd
 	SetEX(ctx context.Context, key string, value any, expiration time.Duration) *StatusCmd
@@ -1086,6 +1087,57 @@ func (c *Compat) MSetNX(ctx context.Context, values ...any) *BoolCmd {
 
 	resp := c.client.Do(ctx, cmd)
 	return newBoolCmd(resp)
+}
+
+// MSetEX sets the given keys to their respective values with expiration options.
+// Supported expiration modes: EX (seconds), PX (milliseconds), EXAT (Unix timestamp in seconds),
+// PXAT (Unix timestamp in milliseconds), or KEEPTTL (preserve existing TTL).
+// Conditions: NX (only if keys don't exist) or XX (only if keys do exist).
+//
+// Returns 1 if all keys were successfully set, 0 if the condition was not satisfied.
+
+func (c *Compat) MSetEX(ctx context.Context, args MSetEXArgs, values ...any) *IntCmd {
+	expandedArgs := argsToSlice(values)
+
+	// Validate even number of key-value pairs
+	if len(expandedArgs)%2 != 0 {
+		errCmd := &IntCmd{}
+		errCmd.SetErr(fmt.Errorf("MSETEX requires an even number of key/value arguments, got %d", len(expandedArgs)))
+		return errCmd
+	}
+
+	numkeys := len(expandedArgs) / 2
+
+	cmd := c.client.B().Arbitrary("MSETEX").Args(strconv.Itoa(numkeys))
+
+	// Add all key-value pairs
+	for i := 0; i < len(expandedArgs); i++ {
+		cmd = cmd.Args(str(expandedArgs[i]))
+	}
+
+	// Add condition (NX or XX)
+	if args.Condition != "" {
+		cmd = cmd.Args(string(args.Condition))
+	}
+
+	// Add expiration options
+	if args.Expiration != nil {
+		switch args.Expiration.Mode {
+		case EX:
+			cmd = cmd.Args("EX", strconv.FormatInt(args.Expiration.Value, 10))
+		case PX:
+			cmd = cmd.Args("PX", strconv.FormatInt(args.Expiration.Value, 10))
+		case EXAT:
+			cmd = cmd.Args("EXAT", strconv.FormatInt(args.Expiration.Value, 10))
+		case PXAT:
+			cmd = cmd.Args("PXAT", strconv.FormatInt(args.Expiration.Value, 10))
+		case KEEPTTL:
+			cmd = cmd.Args("KEEPTTL")
+		}
+	}
+
+	resp := c.client.Do(ctx, cmd.Build())
+	return newIntCmd(resp)
 }
 
 // Set key value [expiration]
--- a/valkeycompat/adapter.go
+++ b/valkeycompat/adapter.go
@@ -1099,6 +1099,13 @@ func (c *Compat) MSetNX(ctx context.Context, values ...any) *BoolCmd {
 func (c *Compat) MSetEX(ctx context.Context, args MSetEXArgs, values ...any) *IntCmd {
 	expandedArgs := argsToSlice(values)
 
+	// Validate there is at least one key/value pair
+	if len(expandedArgs) == 0 {
+		errCmd := &IntCmd{}
+		errCmd.SetErr(fmt.Errorf("MSETEX requires at least one key/value pair: %w", ErrLocalValidation))
+		return errCmd
+	}
+
 	// Validate even number of key-value pairs
 	if len(expandedArgs)%2 != 0 {
 		errCmd := &IntCmd{}
--- a/valkeycompat/adapter.go
+++ b/valkeycompat/adapter.go
@@ -1102,7 +1102,7 @@ func (c *Compat) MSetEX(ctx context.Context, args MSetEXArgs, values ...any) *In
 	// Validate even number of key-value pairs
 	if len(expandedArgs)%2 != 0 {
 		errCmd := &IntCmd{}
-		errCmd.SetErr(fmt.Errorf("MSETEX requires an even number of key/value arguments, got %d", len(expandedArgs)))
+		errCmd.SetErr(fmt.Errorf("MSETEX requires an even number of key/value arguments, got %d: %w", len(expandedArgs), ErrLocalValidation))
 		return errCmd
 	}
 
@@ -1110,9 +1110,26 @@ func (c *Compat) MSetEX(ctx context.Context, args MSetEXArgs, values ...any) *In
 
 	cmd := c.client.B().Arbitrary("MSETEX").Args(strconv.Itoa(numkeys))
 
-	// Add all key-value pairs
-	for i := 0; i < len(expandedArgs); i++ {
-		cmd = cmd.Args(str(expandedArgs[i]))
+	// Validate expiration mode early: if an Expiration was provided it must be a
+	// known, non-empty ExpirationMode. Reject unknown or empty modes so callers
+	// asking for an expiration don't accidentally create persistent keys.
+	if args.Expiration != nil {
+		switch args.Expiration.Mode {
+		case EX, PX, EXAT, PXAT, KEEPTTL:
+			// valid
+		default:
+			errCmd := &IntCmd{}
+			errCmd.SetErr(fmt.Errorf("MSETEX: invalid expiration mode: %q: %w", args.Expiration.Mode, ErrLocalValidation))
+			return errCmd
+		}
+	}
+
+	// Add all key-value pairs and register keys for proper routing in cluster mode.
+	// Use Keys(...) to mark key positions and Args(...) for the corresponding values.
+	for i := 0; i < len(expandedArgs); i += 2 {
+		keyArg := str(expandedArgs[i])
+		valArg := str(expandedArgs[i+1])
+		cmd = cmd.Keys(keyArg).Args(valArg)
 	}
 
 	// Add condition (NX or XX)
--- a/valkeycompat/adapter.go
+++ b/valkeycompat/adapter.go
@@ -1134,9 +1134,7 @@ func (c *Compat) MSetEX(ctx context.Context, args MSetEXArgs, values ...any) *In
 	// Add all key-value pairs and register keys for proper routing in cluster mode.
 	// Use Keys(...) to mark key positions and Args(...) for the corresponding values.
 	for i := 0; i < len(expandedArgs); i += 2 {
-		keyArg := str(expandedArgs[i])
-		valArg := str(expandedArgs[i+1])
-		cmd = cmd.Keys(keyArg).Args(valArg)
+		cmd = cmd.Keys(expandedArgs[i]).Args(expandedArgs[i+1])
 	}
 
 	// Add condition (NX or XX)
--- a/valkeycompat/adapter.go
+++ b/valkeycompat/adapter.go
@@ -1122,7 +1122,7 @@ func (c *Compat) MSetEX(ctx context.Context, args MSetEXArgs, values ...any) *In
 	// asking for an expiration don't accidentally create persistent keys.
 	if args.Expiration != nil {
 		switch args.Expiration.Mode {
-		case EX, PX, EXAT, PXAT, KEEPTTL:
+		case ExpirationEX, ExpirationPX, ExpirationEXAT, ExpirationPXAT, ExpirationKEEPTTL:
 			// valid
 		default:
 			errCmd := &IntCmd{}
@@ -1140,22 +1140,29 @@ func (c *Compat) MSetEX(ctx context.Context, args MSetEXArgs, values ...any) *In
 	}
 
 	// Add condition (NX or XX)
-	if args.Condition != "" {
+	switch args.Condition {
+	case ConditionNX, ConditionXX:
 		cmd = cmd.Args(string(args.Condition))
+	case "":
+		// valid: no condition specified
+	default:
+		errCmd := &IntCmd{}
+		errCmd.SetErr(fmt.Errorf("MSETEX: invalid condition: %q: %w", args.Condition, ErrLocalValidation))
+		return errCmd
 	}
 
 	// Add expiration options
 	if args.Expiration != nil {
 		switch args.Expiration.Mode {
-		case EX:
+		case ExpirationEX:
 			cmd = cmd.Args("EX", strconv.FormatInt(args.Expiration.Value, 10))
-		case PX:
+		case ExpirationPX:
 			cmd = cmd.Args("PX", strconv.FormatInt(args.Expiration.Value, 10))
-		case EXAT:
+		case ExpirationEXAT:
 			cmd = cmd.Args("EXAT", strconv.FormatInt(args.Expiration.Value, 10))
-		case PXAT:
+		case ExpirationPXAT:
 			cmd = cmd.Args("PXAT", strconv.FormatInt(args.Expiration.Value, 10))
-		case KEEPTTL:
+		case ExpirationKEEPTTL:
 			cmd = cmd.Args("KEEPTTL")
 		}
 	}
--- a/valkeycompat/adapter_test.go
+++ b/valkeycompat/adapter_test.go
@@ -1681,6 +1681,337 @@ func testAdapter(resp3 bool) {
 			Expect(mSetNX.Err()).NotTo(HaveOccurred())
 			Expect(mSetNX.Val()).To(Equal(true))
 		})
+
+		It("should MSetEX", func() {
+			// Test MSetEX with EX (seconds)
+			// Returns 1 when all keys are successfully set
+			mSetEX := adapter.MSetEX(ctx, MSetEXArgs{
+				Expiration: &ExpirationOption{
+					Mode:  EX,
+					Value: 10,
+				},
+			}, "key1", "hello1", "key2", "hello2")
+			Expect(mSetEX.Err()).NotTo(HaveOccurred())
+			Expect(mSetEX.Val()).To(Equal(int64(1)))
+			// Verify values were set
+			val1, err := adapter.Get(ctx, "key1").Result()
+			Expect(err).NotTo(HaveOccurred())
+			Expect(val1).To(Equal("hello1"))
+
+			val2, err := adapter.Get(ctx, "key2").Result()
+			Expect(err).NotTo(HaveOccurred())
+			Expect(val2).To(Equal("hello2"))
+
+			// Verify TTL was set
+			ttl1, err := adapter.TTL(ctx, "key1").Result()
+			Expect(err).NotTo(HaveOccurred())
+			Expect(ttl1).To(BeNumerically(">", 0))
+			Expect(ttl1).To(BeNumerically("<=", 10*time.Second))
+
+			// Clean up
+			adapter.Del(ctx, "key1", "key2")
+
+			// Test MSetEX with NX condition
+			// Returns 1 when condition is satisfied and keys are set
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: valkeycompat/command_test.go [hunk 1]
```diff
@@ -612,6 +613,20 @@ var _ = Describe("Commands", func() {
 			Expect(err1).To(Equal(err))
 			Expect(cmd.Err()).To(Equal(err))
 		}
+
+		It("builds MSETEX command in pipeline", func() {
```
dev-sumitGupta: This `It(...)` spec is currently defined inside the `It("Setter & Getter", func() { ... })` callback (which started on line 49). In Ginkgo v2, calling `It` inside another running `It` causes Ginkgo to panic at runtime with:
> `You may not call It() inside another It()`

We need to close the `It("Setter & Getter")` block first so that this test runs as its own standalone spec inside `Describe("Commands", ...)`.

```suggestion
	})

	It("builds MSETEX command in pipeline", func() {
		ctrl := gomock.NewController(GinkgoT())
		defer ctrl.Finish()
		m := mock.NewClient(ctrl)
		p := newPipeline(m)
		// call MSetEX with one key/value and EX 1
		p.MSetEX(ctx, MSetEXArgs{Expiration: &ExpirationOption{Mode: EX, Value: 1}}, "k1", "v1")
		cmds := p.comp.client.(*proxy).cmds
		Expect(len(cmds)).To(BeNumerically(">", 0))
		last := cmds[len(cmds)-1].Commands()
		expected := []string{"MSETEX", "1", "k1", "v1", "EX", "1"}
		Expect(last).To(Equal(expected))
	})
```

valkeycompat/pipeline.go
```diff
@@ -432,6 +432,14 @@ func (c *Pipeline) MSetNX(ctx context.Context, values ...any) *BoolCmd {
 	return ret
 }
 
+func (c *Pipeline) MSetEX(ctx context.Context, args MSetEXArgs, values ...any) *IntCmd {
```
dev-sumitGupta, line 435:
```go
func (c *Pipeline) MSetEX(ctx context.Context, args MSetEXArgs, values ...any) *IntCmd {
	ret := c.comp.MSetEX(ctx, args, values...)
	if !errors.Is(ret.Err(), ErrLocalValidation) {
		c.rets = append(c.rets, ret)
	}
	return ret
}
```

Skipping `c.rets = append(c.rets, ret)` when `ret.Err()` contains `ErrLocalValidation` breaks standard pipeline behavior:

1. **Commands disappear from `pipe.Exec()` results:**
   If a caller queues `cmd1 := pipe.MSetEX(...)` (which fails validation) followed by `cmd2 := pipe.Get(...)`, `cmd1` is not added to `c.rets`. When `pipe.Exec(ctx)` is called, the returned `[]Cmder` slice only has length 1 (containing only `cmd2`). `cmd1` completely vanishes from the execution results.

2. **Errors are swallowed if it is the only command:**
   If a pipeline only runs an invalid `MSetEX`, nothing is added to `p.cmds`. When `pipe.Exec(ctx)` runs:
   ```go
   if len(p.cmds) == 0 {
       return nil, nil
   }
   ```
   `pipe.Exec(ctx)` returns `(nil, nil)`, so the caller gets no error and falsely assumes the command succeeded.

3. **Inconsistency with adjacent pipeline methods:**
   Notice how `MSetNX` (line 430) and `Set` (line 444) are implemented:
   ```go
   func (c *Pipeline) MSetNX(ctx context.Context, values ...any) *BoolCmd {
       ret := c.comp.MSetNX(ctx, values...)
       c.rets = append(c.rets, ret)
       return ret
   }
   ```
   All other pipeline methods append to `c.rets` unconditionally and rely on the server to reject bad arguments during `Exec()`.

#### Recommendation:
Align `MSetEX` with `MSet` / `MSetNX`: remove client-side early returns from `Compat.MSetEX` and allow the command to be sent through `c.client.Do()` so `p.cmds` and `c.rets` stay 1-to-1 in sync, and validation errors are returned natively by the server on `pipe.Exec()`.

valkeycompat/adapter.go
```diff
@@ -1130,6 +1131,81 @@ func (c *Compat) MSetNX(ctx context.Context, values ...any) *BoolCmd {
...
+			errCmd := &IntCmd{}
+			errCmd.SetErr(fmt.Errorf("MSETEX: invalid expiration mode: %q: %w", args.Expiration.Mode, ErrLocalValidation))
+			return errCmd
+		}
+	}
+
+	// Add all key-value pairs and register keys for proper routing in cluster mode.
+	// Use Keys(...) to mark key positions and Args(...) for the corresponding values.
+	for i := 0; i < len(expandedArgs); i += 2 {
+		keyArg := str(expandedArgs[i])
```
dev-sumitGupta: `expandedArgs` is already a `[]string` (returned by `argsToSlice`), so calling `str()` is unnecessary and causes extra `any` interface boxing allocations. We can pass the slice elements directly:

```diff
 	for i := 0; i < len(expandedArgs); i += 2 {
-		keyArg := str(expandedArgs[i])
-		valArg := str(expandedArgs[i+1])
-		cmd = cmd.Keys(keyArg).Args(valArg)
+		cmd = cmd.Keys(expandedArgs[i]).Args(expandedArgs[i+1])
 	}
```

valkeycompat/adapter.go
```diff
@@ -1130,6 +1131,81 @@ func (c *Compat) MSetNX(ctx context.Context, values ...any) *BoolCmd {
...
+		valArg := str(expandedArgs[i+1])
+		cmd = cmd.Keys(keyArg).Args(valArg)
+	}
+
+	// Add condition (NX or XX)
+	if args.Condition != "" {
```
dev-sumitGupta: 💡 Optional Optimization: Fail fast on invalid `args.Condition`

Currently, if an invalid condition is provided, it gets sent over the wire and the Valkey server rejects it with a syntax error.

As an optional fail-fast optimization (matching how `args.Expiration.Mode` is handled above), we can validate `args.Condition` locally to avoid an unnecessary network round-trip on invalid input:

```go
switch args.Condition {
case NX, XX:
	cmd = cmd.Args(string(args.Condition))
case "":
	// valid: no condition specified
default:
	errCmd := &IntCmd{}
	errCmd.SetErr(fmt.Errorf("MSETEX: invalid condition: %q: %w", args.Condition, ErrLocalValidation))
	return errCmd
}
```

valkeycompat/adapter_test.go
```diff
@@ -1805,6 +1806,337 @@ func testAdapter(resp3 bool) {
...
+				},
+			}, "compare1", "value1", "compare2", "value2")
+
+			Expect(result.Err()).NotTo(HaveOccurred())
+			Expect(result.Val()).To(Equal(int64(1)))
+
+			pttl1 := adapter.PTTL(ctx, "compare1").Val()
+			pttl2 := adapter.PTTL(ctx, "compare2").Val()
+
+			fmt.Printf("valkeycompat PTTL compare1 = %d ms\n", pttl1.Milliseconds())
```
dev-sumitGupta: 1. **Remove leftover `fmt.Printf` calls (lines 2098–2099):**
   Unit/integration tests should not write debugging logs to standard output.

2. **Reduce `Eventually` wait time (lines 2104–2111):**
   Because this test runs for **both RESP2 and RESP3**, waiting 12 seconds for keys to expire adds ~24 seconds of idle wait time to every test run. 

   Using a 1-second TTL with a 2-second `Eventually` window (matching the `It("should SetEX")` test in this same file) verifies expiration just as effectively while keeping CI fast:

```diff
 	result := adapter.MSetEX(ctx, MSetEXArgs{
 		Expiration: &ExpirationOption{
 			Mode:  EX,
-			Value: 10,
+			Value: 1,
 		},
 	}, "compare1", "value1", "compare2", "value2")

 	Expect(result.Err()).NotTo(HaveOccurred())
 	Expect(result.Val()).To(Equal(int64(1)))

 	pttl1 := adapter.PTTL(ctx, "compare1").Val()
 	pttl2 := adapter.PTTL(ctx, "compare2").Val()

-	fmt.Printf("valkeycompat PTTL compare1 = %d ms\n", pttl1.Milliseconds())
-	fmt.Printf("valkeycompat PTTL compare2 = %d ms\n", pttl2.Milliseconds())

 	Expect(pttl1).To(BeNumerically(">", 0))
 	Expect(pttl2).To(BeNumerically(">", 0))

 	Eventually(func() int64 {
 		return adapter.Exists(ctx, "compare1").Val()
-	}, 12*time.Second, 100*time.Millisecond).Should(Equal(int64(0)))
+	}, 2*time.Second, 50*time.Millisecond).Should(Equal(int64(0)))

 	Eventually(func() int64 {
 		return adapter.Exists(ctx, "compare2").Val()
-	}, 12*time.Second, 100*time.Millisecond).Should(Equal(int64(0)))
+	}, 2*time.Second, 50*time.Millisecond).Should(Equal(int64(0)))
```

valkeycompat/command.go
```diff
@@ -2359,6 +2362,44 @@ type SetArgs struct {
...
+	// EX sets expiration in seconds
+	EX ExpirationMode = "EX"
+	// PX sets expiration in milliseconds
+	PX ExpirationMode = "PX"
+	// EXAT sets expiration as Unix timestamp in seconds
+	EXAT ExpirationMode = "EXAT"
+	// PXAT sets expiration as Unix timestamp in milliseconds
+	PXAT ExpirationMode = "PXAT"
+	// KEEPTTL keeps the existing TTL
+	KEEPTTL ExpirationMode = "KEEPTTL"
```
dev-sumitGupta: Package `valkeycompat` already defines an exported constant in `adapter.go:48`:
```go
const KeepTTL = -1
```

Adding `KEEPTTL` here (line 2386):
```go
const (
	...
	KEEPTTL ExpirationMode = "KEEPTTL"
)
```
creates two identically named constants in the same package differing only by casing (`valkeycompat.KeepTTL` vs `valkeycompat.KEEPTTL`), which can lead to confusion and autocomplete collisions for users of the package.

Additionally, exporting bare names like `EX`, `PX`, `NX`, `XX` directly in package `valkeycompat` pollutes the package namespace.

#### Recommendation:
Consider prefixing these constants (similar to how `HSetEX` names its constants in this same file at line 2327: `HSetEXExpirationEX`, `HSetEXExpirationKEEPTTL`):

```go
type ExpirationMode string

const (
	ExpirationEX      ExpirationMode = "EX"
	ExpirationPX      ExpirationMode = "PX"
	ExpirationEXAT    ExpirationMode = "EXAT"
	ExpirationPXAT    ExpirationMode = "PXAT"
	ExpirationKEEPTTL ExpirationMode = "KEEPTTL"
)
```
This avoids collision with `valkeycompat.KeepTTL` and keeps the package namespace clear.

valkeycompat/command_test.go
sonalijhala-23: Changes have been incorporated.

valkeycompat/pipeline.go
sonalijhala-23: Changes have been incorporated.

valkeycompat/adapter_test.go
sonalijhala-23: Changes have been incorporated.

valkeycompat/command.go
sonalijhala-23: Changes have been incorporated.

valkeycompat/adapter.go
sonalijhala-23: Changes have been incorporated.

dev-sumitGupta:
@sonalijhala-23 can you please look into the failing tests?

dev-sumitGupta (approved):

valkeycompat/command.go
```diff
@@ -5686,4 +5727,4 @@ func (cmd *FunctionStatsCmd) parseRunningScripts(msg valkey.ValkeyMessage) ([]Ru
 
 	}
 	return vals, err
-}
+}
```
rueian:
```suggestion
}

```

valkeycompat/adapter.go
```diff
@@ -6808,4 +6889,4 @@ func isEmptyValue(v reflect.Value) bool {
 		return v.IsNil()
 	}
 	return false
-}
+}
```
rueian:
```suggestion
}

```

valkeycompat/command.go
```diff
@@ -2359,6 +2362,44 @@ type SetArgs struct {
...
+	// ExpirationEX sets expiration in seconds
+	ExpirationEX ExpirationMode = "EX"
+	// ExpirationPX sets expiration in milliseconds
+	ExpirationPX ExpirationMode = "PX"
+	// ExpirationEXAT sets expiration as Unix timestamp in seconds
+	ExpirationEXAT ExpirationMode = "EXAT"
+	// ExpirationPXAT sets expiration as Unix timestamp in milliseconds
+	ExpirationPXAT ExpirationMode = "PXAT"
+	// ExpirationKEEPTTL keeps the existing TTL
+	ExpirationKEEPTTL ExpirationMode = "KEEPTTL"
```
rueian: Do these constants have the same names as go-redis? go-redis seems don't have Condition and Expiration prefixes.

valkeycompat/command.go
```diff
@@ -39,6 +39,9 @@ import (
 	"github.com/valkey-io/valkey-go/internal/util"
 )
 
+
+var ErrLocalValidation = errors.New("valkeycompat: local validation error")
```
rueian: Does go-redis have this error?

valkeycompat/command.go
sonalijhala-23: Right, in go-redis they are named NX, XX for SetCondition and EX, PX, EXAT, PXAT, KEEPTTL, PERSIST for ExpirationMode without any prefixes. I have updated the constant names to match go-redis exactly.

valkeycompat/command.go
sonalijhala-23: No, go-redis does not have ErrLocalValidation. In go-redis, MSetEX passes the arguments straight to the server and lets Redis/Valkey return the syntax/argument error, rather than doing client-side validation. I have removed ErrLocalValidation and aligned the implementation with go-redis.

valkeycompat/adapter_test.go
```diff
@@ -13596,6 +13945,7 @@ func testAdapterSearchRESP2() {
 				"description": "Health trends for the new year, including fitness regimes.",
 				"rating":      4,
 			})
+			WaitForIndexing(client, "idx1", 2)
```
rueian: why do you change this?

valkeycompat/adapter_test.go
```diff
@@ -13673,7 +14023,6 @@ func testAdapterSearchRESP2() {
 			val, err := adapter.FTCreate(ctx, "idx1", &FTCreateOptions{}, &FieldSchema{FieldName: "geom", FieldType: SearchFieldTypeGeoShape, GeoShapeFieldType: "FLAT"}).Result()
 			Expect(err).NotTo(HaveOccurred())
 			Expect(val).To(BeEquivalentTo("OK"))
-			WaitForIndexing(client, "idx1", 2)
```
rueian, line 13676: why do you change this?

valkeycompat/command_test.go
```diff
@@ -613,6 +614,20 @@ var _ = Describe("Commands", func() {
...
+		m := mock.NewClient(ctrl)
+		p := newPipeline(m)
+		// call MSetEX with one key/value and EX 1
+		p.MSetEX(ctx, MSetEXArgs{Expiration: &ExpirationOption{Mode: EX, Value: 1}}, "k1", "v1")
+		cmds := p.comp.client.(*proxy).cmds
+		Expect(len(cmds)).To(BeNumerically(">", 0))
+		last := cmds[len(cmds)-1].Commands()
+		expected := []string{"MSETEX", "1", "k1", "v1", "EX", "1"}
+		Expect(last).To(Equal(expected))
+	})
```
rueian: this test is not needed as it is tested in the pipeline_test.go

valkeycompat/pipeline_test.go
```diff
@@ -738,8 +739,9 @@ var golden = `[
     ["INCRBY","1","1"],
     ["INCRBYFLOAT","1","1"],
     ["MGET","1","2"],
-    ["MSET","1","2"],
-    ["MSETNX","1","2"],
+	["MSET","1","2"],
+	["MSETNX","1","2"],
+	["MSETEX","1","1","2","EX","1"],
```
rueian: Do indentations here align with other lines?

valkeycompat/adapter_test.go
sonalijhala-23: Apologies for that! This was an unintended change introduced while resolving merge conflicts . I will revert this change.

valkeycompat/adapter_test.go
sonalijhala-23: his was an unintended change introduced while resolving merge conflicts . I will revert this change.

valkeycompat/pipeline_test.go
sonalijhala-23: Changes incorporated

valkeycompat/command_test.go
sonalijhala-23: Changes incorporated

valkeycompat/adapter.go
```diff
@@ -1133,6 +1134,58 @@ func (c *Compat) MSetNX(ctx context.Context, values ...any) *BoolCmd {
...
+			cmd = cmd.Args("PX", strconv.FormatInt(args.Expiration.Value, 10))
+		case EXAT:
+			cmd = cmd.Args("EXAT", strconv.FormatInt(args.Expiration.Value, 10))
+		case PXAT:
+			cmd = cmd.Args("PXAT", strconv.FormatInt(args.Expiration.Value, 10))
+		case KEEPTTL:
+			cmd = cmd.Args("KEEPTTL")
+		default:
+			if args.Expiration.Mode != "" {
+				cmd = cmd.Args(string(args.Expiration.Mode), strconv.FormatInt(args.Expiration.Value, 10))
```
rueian: Is this correct? EX, PX, EXAT, PXT, KEEPTTL are the only options the server accepts. Sending anything else is an error.

sonalijhala-23: Initially, this default case was added when removing client-side validation, with the thought of passing any invalid mode through to let the server return a syntax error. I'll remove the default block so we only handle the supported expiration modes .

rueian (approved):

PR #165, opened 2026-09-02
feat(valkeycompat): add SetFromBuffer for go-redis parity

kumariam-ai:
****PR Description:** ## Summary**

Adds ⁠ SetFromBuffer ⁠ to ⁠ valkeycompat ⁠ so the go-redis adapter matches
go-redis v9.21+.

⁠ Compat.Set ⁠ already accepts ⁠ []byte ⁠, but ⁠ str() ⁠ copies. This path uses
⁠ valkey.BinaryString(buf) ⁠ and issues a plain ⁠ SET key value ⁠ with no TTL.

**## API**
SetFromBuffer(ctx context.Context, key string, buf []byte) *StatusCmd

**Signed-off-by**: Amrita Kumari kumariam@google.com
**Co-authored-by**: Omkar Mestry omanges@google.com

### DIFF
--- a/valkeycompat/adapter_test.go
+++ b/valkeycompat/adapter_test.go
@@ -27,14 +27,21 @@
 package valkeycompat
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"fmt"
 	"math"
+	"net"
 	"strconv"
 	"strings"
 	"testing"
 	"time"
+
+	"errors"
+
+	"github.com/valkey-io/valkey-go/mock"
+	"go.uber.org/mock/gomock"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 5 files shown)


valkeycompat/adapter_test.go
```diff
@@ -27,15 +27,22 @@
...
+	"net"
 	"strconv"
 	"strings"
 	"testing"
 	"time"
 
+	"errors"
+
+	"github.com/valkey-io/valkey-go/mock"
+	"go.uber.org/mock/gomock"
```
rueian: Why not just follow the existing test for SET? I think we don't need to have new imports for this.

valkeycompat/pipeline.go
```diff
@@ -464,6 +464,12 @@ func (c *Pipeline) SetRange(ctx context.Context, key string, offset int64, value
 	return ret
 }
 
+// SetFromBuffer is not supported for Pipeline because it performs a zero-copy
+// conversion of the provided buffer to string which cannot be safely deferred.
+func (c *Pipeline) SetFromBuffer(ctx context.Context, key string, buf []byte) *StatusCmd {
+	panic("SetFromBuffer is not supported in Pipeline (zero-copy buffer conversion)")
```
rueian: We can support SetFromBuffer in the pipeline because the value is actually not deferred.

valkeycompat/tx.go
```diff
@@ -147,3 +147,10 @@ func (t *tx) Close(_ context.Context) error {
 	t.cancel()
 	return nil
 }
+
+// SetFromBuffer is intentionally unsupported in TxPipeline for the same
+// reasons as Pipeline.SetFromBuffer: zero-copy buffer conversions cannot be
+// safely deferred. Panic to make misuse explicit.
+func (c *TxPipeline) SetFromBuffer(ctx context.Context, key string, buf []byte) *StatusCmd {
+	panic("SetFromBuffer is not supported in TxPipeline (zero-copy buffer conversion)")
```
rueian: Same here. We can support SetFromBuffer in the pipeline because the value is actually not deferred.

valkeycompat/adapter_test.go
kumariam-ai: Thank you for your review, updated the code as per suggestion

valkeycompat/pipeline.go
kumariam-ai: Updated

valkeycompat/tx.go
kumariam-ai: updated

rueian (approved):

PR #170, opened 2026-09-04
provide ser/de funcs with context and cache key

nelzblooket:
I have used this construct in my work projects for doing things like talking to telemetry and/or memoizing the deserialization with an in-memory object cache.

### DIFF
--- a/valkeyaside/typed_aside.go
+++ b/valkeyaside/typed_aside.go
@@ -17,8 +17,8 @@ type TypedCacheAsideClient[T any] interface {
 // It provides a typed cache-aside client that allows caching and retrieving values of a specific type T.
 type typedCacheAsideClient[T any] struct {
 	client       CacheAsideClient
-	serializer   func(*T) (string, error)
-	deserializer func(string) (*T, error)
+	serializer   func(context.Context, string, *T) (string, error)
+	deserializer func(context.Context, string, string) (*T, error)
 }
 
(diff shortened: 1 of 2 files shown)


valkeyaside/typed_aside.go
```diff
@@ -29,6 +29,26 @@ func NewTypedCacheAsideClient[T any](
...
+			return deserializer(str)
+		},
+	}
+}
+
+// NewTypedCacheAsideClientWithContext creates a new TypedCacheAsideClient instance that provides a typed cache-aside client.
+// The client, serializer, and deserializer functions are used to interact with the underlying cache.
+// The serializer function is used to convert the provided value of type T to a string, and the deserializer function
+// is used to convert the cached string value back to the original type T.
+func NewTypedCacheAsideClientWithContext[T any](
```
rueian: The naming of `NewTypedCacheAsideClientWithContext` is not good. Let's do this:

```go
// TypedCodec converts values of type T to and from their cached string form.
// Both functions receive the context and cache key of the in-flight Get.
type TypedCodec[T any] struct {
	Marshal   func(ctx context.Context, key string, val *T) (string, error)
	Unmarshal func(ctx context.Context, key string, str string) (*T, error)
}

func NewTypedCacheAsideClientWithCodec[T any](
	client CacheAsideClient,
	codec TypedCodec[T],
) TypedCacheAsideClient[T]
```

rueian: Hi @nelzblooket, please also sign off your commit to make DCO check pass.

rueian (approved):

PR #181, opened 2026-09-12
feat(cluster): add opt-in PreferClusterShards to use CLUSTER SHARDS before version 8

DhruvilK7:
By default the client refreshes cluster topology with `CLUSTER SHARDS` only on servers version 8 and above, and uses `CLUSTER SLOTS` below that. This was done because `CLUSTER SHARDS` had a server side bug on older versions.

The bug is fixed in Valkey `7.2.6` and Redis `8.0`, so users on those versions can now opt in to `CLUSTER SHARDS` earlier by setting PreferClusterShards. This keeps the health field from `CLUSTER SHARDS` during rolling engine upgrades, which avoids routing commands to nodes that are still loading.

The option is off by default so existing behavior does not change. It only takes effect on servers version `7` and above, so it never sends `CLUSTER SHARDS` to a server that does not support the command.

### DIFF
--- a/cluster.go
+++ b/cluster.go
@@ -198,19 +198,19 @@ func clusterRefreshAutoMaxDelay(n int) time.Duration {
 }
 
 type clusterslots struct {
-	addr  string
-	reply ValkeyResult
-	ver   int
+	addr      string
+	reply     ValkeyResult
+	useShards bool
 }
 
 func (s clusterslots) parse(tls bool) map[string]group {
-	if s.ver < 8 {
+	if !s.useShards {
 		return parseSlots(s.reply.val, s.addr)
 	}
 	return parseShards(s.reply.val, s.addr, tls)
 }
 
-func getClusterSlots(c conn, timeout time.Duration) clusterslots {
+func getClusterSlots(c conn, timeout time.Duration, preferShards bool) clusterslots {
 	var ctx context.Context
 	var cancel context.CancelFunc
 	if timeout > 0 {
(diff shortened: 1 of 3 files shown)


cluster.go
```diff
@@ -429,14 +385,15 @@ func (c *clusterClient) clusterRefreshConns() []conn {
...
 func (c *clusterClient) refreshConns(pending []conn, batchDelay time.Duration) (result clusterslots, err error) {
 	results := make(chan clusterslots, len(pending))
+	preferShards := c.opt.ClusterOption.PreferClusterShards
 	for i := 0; i < len(pending); i++ {
 		if i&3 == 0 { // batch CLUSTER SLOTS/CLUSTER SHARDS for every 4 connections
 			if i > 0 && batchDelay > 0 {
 				time.Sleep(batchDelay)
 			}
 			for j := i; j < i+4 && j < len(pending); j++ {
 				go func(c conn, timeout time.Duration) {
```
```suggestion
				go func(c conn, timeout time.Duration, preferShards bool) {
```

cluster.go
```diff
@@ -429,14 +385,15 @@ func (c *clusterClient) clusterRefreshConns() []conn {
...
-					results <- getClusterSlots(c, timeout)
+					results <- getClusterSlots(c, timeout, preferShards)
 				}(pending[j], c.opt.ConnWriteTimeout)
```
```suggestion
				}(pending[j], c.opt.ConnWriteTimeout, preferShards)
```

rueian (approved):

PR #185, opened 2026-09-17
(valkeyaside) demonstrate a deadlock

nelzblooket:
I stumbled upon this problem when creating self-cleaning tooling for running integration tests that uses `valkeyaside`.


nelzblooket:
If it helps at all, Claude suggested I add this code to my tooling, and it seems to work for the short-term. (I don't know if it is helpful for the long term.)

```
client, err := valkeyaside.NewClient(valkeyaside.ClientOption{
    ClientOption: valkey.ClientOption{InitAddress: addrs, SelectDB: 0},
    ClientBuilder: func(opt valkey.ClientOption) (valkey.Client, error) {
        orig := opt.OnInvalidations
        opt.OnInvalidations = func(m []valkey.ValkeyMessage) {
            if m == nil {
                go orig(nil) // avoid blocking the read loop on the DEL
                return
            }
            orig(m)
        }
        return valkey.NewClient(opt)
    },
})
```

rueian:
> If it helps at all, Claude suggested I add this code to my tooling, and it seems to work for the short-term. (I don't know if it is helpful for the long term.)
> 
> ```
> client, err := valkeyaside.NewClient(valkeyaside.ClientOption{
>     ClientOption: valkey.ClientOption{InitAddress: addrs, SelectDB: 0},
>     ClientBuilder: func(opt valkey.ClientOption) (valkey.Client, error) {
>         orig := opt.OnInvalidations
>         opt.OnInvalidations = func(m []valkey.ValkeyMessage) {
>             if m == nil {
>                 go orig(nil) // avoid blocking the read loop on the DEL
>                 return
>             }
>             orig(m)
>         }
>         return valkey.NewClient(opt)
>     },
> })
> ```

Hi @nelzblooket, thanks for the report. Would you like to update the PR to include the fix by putting the DEL in another goroutine?

rueian (approved):

